PR #357, opened 2026-04-13
fix(mt#673): Align CredentialSource values between credential-resolver and config-validator

edobry:
## Summary

- `CredentialSource` type changed from `"environment" | "file" | "prompt"` to `"env" | "file" | "keychain" | "manual"` — now matches the values `config-validator.ts` accepts from user config files
- Updated `isCredentialConfig()` guard, `resolveCredentialFromConfig()` switch cases, and the Zod `credentialSourceSchema` in `schemas/base.ts`
- Updated `CredentialSource` in `types.ts` and `getTokenSource()` return type in `schemas/github.ts`
- Added `credential-resolver.test.ts` with 8 tests covering all four source values (`env`, `file`, `keychain`, `manual`)

Previously, `source: "env"` would pass validation but silently fall through the resolver's switch (hitting `default: return undefined`). Now the two layers are consistent.

## Test plan
- [ ] New tests in `credential-resolver.test.ts` all pass (`bun test src/domain/configuration/credential-resolver.test.ts`)
- [ ] No TypeScript errors (`bun run typecheck`)
- [ ] Config with `source: "env"` now resolves correctly instead of returning undefined

### DIFF
--- a/src/domain/configuration/credential-resolver.test.ts
+++ b/src/domain/configuration/credential-resolver.test.ts
@@ -1,0 +1,75 @@
+/**
+ * Tests for DefaultCredentialResolver
+ *
+ * Verifies that all valid CredentialSource values are handled correctly
+ * and that the resolver aligns with config-validator's accepted values.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { DefaultCredentialResolver } from "./credential-resolver";
+import type { CredentialConfig } from "./credential-resolver";
+
+describe("DefaultCredentialResolver.resolveCredentialFromConfig", () => {
+  const resolver = new DefaultCredentialResolver();
+
+  describe('source: "env"', () => {
+    test("returns token when provided", async () => {
+      const config: CredentialConfig = { source: "env", token: "my-token" };
+      const result = await resolver.resolveCredentialFromConfig(config);
+      expect(result).toBe("my-token");
+    });
+
+    test("returns api_key when token is absent", async () => {
+      const config: CredentialConfig = { source: "env", api_key: "my-api-key" };
+      const result = await resolver.resolveCredentialFromConfig(config);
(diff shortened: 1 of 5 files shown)


edobry (commented):
## Review: Credential source value alignment

### Findings

No blocking issues found.

**[NON-BLOCKING]** `credential-resolver.ts:22-23` — Two new optional fields (`env_var`, `path`) added to `CredentialConfig` but never read by the resolver. The `env` case just returns `credentialConfig.token || credentialConfig.api_key` without consulting `env_var`. Low risk since the fields are optional, but they're dead code that may mislead future callers into thinking they work.

### Checked and clear

- `CredentialSource` type aligned across `credential-resolver.ts`, `types.ts`, and `schemas/base.ts` — all now use `["env", "file", "keychain", "manual"]`
- `config-validator.ts` already uses `["env", "file", "keychain", "manual"]` at lines 192 and 277 — confirmed alignment
- `isCredentialConfig` guard updated to match new values
- `getTokenSource` in `schemas/github.ts` updated from `"environment"` to `"env"`
- `keychain` and `manual` cases throw explicit not-implemented errors (better than silent fallthrough)
- Tests cover all 4 source values including error paths

### Spec verification

**Task:** mt#673

| CredentialSource accepts same values as config-validator | Met | Both use `["env", "file", "keychain", "manual"]` |
| Only "file" overlaps — fix the mismatch | Met | All 4 values now shared |
| Config with source: "env" no longer silently fails | Met | `case "env"` resolves token/api_key |
| Add test that catches future drift | Met | 8 tests in `credential-resolver.test.ts` |

edobry:
Closing to recreate with clean session state. Original commit preserved in git history. The replacement PR will include the dead-field fix from code review.

PR #362, opened 2026-04-13
fix(mt#704): Fix session delete silent failure and add workspace cleanup

edobry:
## Summary
- `deleteSessionImpl` now returns structured `DeleteSessionResult` with error messages instead of bare boolean
- `SessionDbAdapter.deleteSession` no longer swallows storage errors -- only returns false for "not found" cases
- Management command includes `error` field in response when `success: false`
- Session delete now removes workspace directory from filesystem before deleting the DB record
- **mt#706**: Session delete no longer uses auto-detection -- requires explicit name or task ID to prevent wrong-session deletion
- **mt#706**: Postgres storage `delete()` no longer swallows errors either

## Context
During testing, a session DB record silently disappeared. Investigation found:
1. `deleteSessionImpl` used `allowAutoDetection: true` -- if called with wrong parameter names, it could resolve to and delete the wrong session
2. Both JSON and Postgres storage delete methods swallowed all errors, returning `false` with no diagnostics
3. Management command returned `{ success: false }` with no error field

## Test plan
- [x] Type check passes (tsc --noEmit clean)
- [x] All 311 session + storage tests pass
- [ ] Manual test: delete with no name/task returns error message about requiring explicit identification
- [ ] Manual test: delete nonexistent session returns error message
- [ ] Manual test: delete existing session removes both DB record and directory

### DIFF
--- a/src/adapters/shared/commands/session/management-commands.ts
+++ b/src/adapters/shared/commands/session/management-commands.ts
@@ -21,7 +21,7 @@ export function createSessionDeleteCommand(deps: SessionCommandDependencies): Co
     execute: withErrorLogging("session.delete", async (params: Record<string, unknown>) => {
       const { deleteSessionFromParams } = await import("../../../../domain/session");
 
-      const deleted = await deleteSessionFromParams({
+      const result = await deleteSessionFromParams({
         name: params.name as string | undefined,
         task: params.task as string | undefined,
         force: (params.force as boolean | undefined) ?? false,
@@ -30,8 +30,9 @@ export function createSessionDeleteCommand(deps: SessionCommandDependencies): Co
       });
 
       return {
-        success: deleted,
+        success: result.deleted,
         session: params.name || params.task,
+        ...(result.error ? { error: result.error } : {}),
       };
     }),
   };
--- a/src/domain/session.ts
+++ b/src/domain/session.ts
@@ -233,7 +233,7 @@ export async function startSessionFromParams(
 export async function deleteSessionFromParams(
   params: SessionDeleteParams,
   depsInput?: { sessionDB?: SessionProviderInterface }
-): Promise<boolean> {
+): Promise<import("./session/session-lifecycle-operations").DeleteSessionResult> {
   const service = new SessionService(
     await resolvePartialDeps({ sessionProvider: depsInput?.sessionDB })
   );
--- a/src/domain/session/session-auto-detection-integration.test.ts
+++ b/src/domain/session/session-auto-detection-integration.test.ts
@@ -114,7 +114,7 @@ describe("Session Command Domain Logic", () => {
         }
       );
 
-      expect(result).toBe(true);
+      expect(result.deleted).toBe(true);
     });
 
     test("deletes session by explicit task ID", async () => {
@@ -129,7 +129,7 @@ describe("Session Command Domain Logic", () => {
         }
       );
 
-      expect(result).toBe(true);
+      expect(result.deleted).toBe(true);
     });
 
     test("throws ResourceNotFoundError for non-existent session", async () => {
@@ -157,7 +157,8 @@ describe("Session Command Domain Logic", () => {
         }
       );
 
-      expect(result).toBe(false);
+      expect(result.deleted).toBe(false);
+      expect(result.error).toBeDefined();
     });
   });
 
@@ -178,7 +179,7 @@ describe("Session Command Domain Logic", () => {
         { task: taskId, force: true, json: false },
         { sessionDB: mockSessionProvider }
       );
-      expect(deleteResult).toBe(true);
+      expect(deleteResult.deleted).toBe(true);
 
       // All commands should have resolved to the same session
     });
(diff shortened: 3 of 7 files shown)


edobry (commented):
## Review: Session delete silent failure fix

**CI status:** Build pending, placeholder test check passing

### Findings

No blocking issues found.

**[NON-BLOCKING]** `session-lifecycle-operations.ts:116-131` — The workspace directory cleanup duplicates logic from `cleanupSessionImpl` (lines 184+), which handles dry-run, safety validation, and multiple directories. The delete path intentionally keeps it simple (single directory, non-fatal on failure), which is reasonable, but a future refactor could extract a shared `removeSessionDirectory` helper to avoid drift.

**[NON-BLOCKING]** Bug 3 from the spec (MCP `isError` flag) is not addressed in this PR. The error message IS now included in the JSON response (visible to agents), but the MCP protocol-level `isError` flag is not set. This is a broader concern affecting all MCP tools, not just delete — reasonable to defer to a separate task.

### Checked and clear

- `DeleteSessionResult` interface properly typed with `deleted: boolean` and optional `error`
- `SessionDbAdapter.deleteSession` no longer wraps in catch-all — storage errors propagate, only "not found" returns `false`
- Management command spreads `error` field into response when present
- `deleteSessionFromParams` return type updated from `boolean` to `DeleteSessionResult`
- `SessionService.delete` return type updated to match
- All 4 test assertions updated from `expect(result).toBe(true)` to `expect(result.deleted).toBe(true)`
- Filesystem cleanup uses `existsSync` + `rmSync` with `recursive: true, force: true` — appropriate for session dirs
- `getSessionsDir()` is the canonical path helper (verified at `utils/paths.ts:53`)

### Spec verification

**Task:** mt#704

| deleteSessionImpl returns structured result with error reason | Met | `DeleteSessionResult` with `error?: string` |
| SessionDbAdapter.deleteSession propagates errors | Met | Catch-all removed, only returns `false` for not-found |
| Management command includes error field when success=false | Met | `management-commands.ts:35` spreads error |
| MCP tool response uses isError: true on failure | Not met | Deferred — broader MCP concern, not delete-specific |
| Session delete cleans up filesystem worktree | Met | `rmSync` before DB delete, non-fatal on failure |
| Tests cover delete scenarios | Met | 4 assertions updated, error case checks `result.error` |

**Action required:** Criterion 4 (MCP `isError`) is deferred. Should create a follow-up task for MCP-wide error signaling rather than blocking this fix. The error message is already visible in the response JSON, which addresses the original silent failure problem.

edobry (commented):
## Review: Session delete fixes (mt#704 + mt#706)

### Findings

**[NON-BLOCKING]** `postgres-storage.ts:408-418` — `deleteEntity()` still has the same error-swallowing pattern (`catch → return false`). The PR correctly fixes `SessionDbAdapter.deleteSession` and the `delete()` wrapper, but the underlying `deleteEntity()` is one layer deeper and still swallows. Additionally, SQL `DELETE WHERE` on a non-existent row succeeds with 0 rows affected, so `deleteEntity` always returns `true` regardless of whether the row existed. This means the `deleted` field in `DeleteSessionResult` is unreliable for Postgres — it always reports `true`. Worth a follow-up to check `rowsAffected` from drizzle.

**[NON-BLOCKING]** `session-lifecycle-operations.ts:116-131` — Directory cleanup logic is simple and correct. Same duplication note as previous review re: `cleanupSessionImpl`, acceptable for now.

### Checked and clear

- `DeleteSessionResult` interface properly typed with `deleted: boolean` and optional `error`
- `SessionDbAdapter.deleteSession` catch-all removed — only `false` for not-found
- `PostgresStorage.delete()` catch-all removed — errors propagate to adapter
- Management command spreads `error` field into response
- Auto-detection disabled in delete path (`allowAutoDetection: false`) — prevents wrong-session deletion
- Explicit guard: returns error if no `name`, `task`, or `repo` provided
- `getSessionsDir()` is the canonical path helper
- All test assertions updated to use structured result
- Session domain tests (298) and storage tests (311 total) all pass

### Spec verification

**Task:** mt#704

| deleteSessionImpl returns structured result with error reason | Met | `DeleteSessionResult` with `error?: string` |
| SessionDbAdapter.deleteSession propagates errors | Met | Catch-all removed |
| Management command includes error field when success=false | Met | `management-commands.ts:35` |
| MCP tool response uses isError: true on failure | Not met | Deferred — broader MCP concern |
| Session delete cleans up filesystem worktree | Met | `rmSync` before DB delete |
| Tests cover delete scenarios | Met | 4 assertions updated + error case |

**Task:** mt#706

| Delete with no name/task/repo returns error | Met | Guard at lifecycle-operations.ts:95 |
| Explicit name or task still works | Met | Tests pass for both paths |
| Other session commands unaffected | Met | Only `deleteSessionImpl` changed |

**Action items for follow-up:**
1. MCP `isError` signaling (deferred, broader concern)
2. `PostgresStorage.deleteEntity` error-swallowing + always-true return (deeper layer fix)

PR #365, opened 2026-04-13
fix(mt#673): Align CredentialSource values between resolver and validator

edobry:
## Summary
- `CredentialSource` type changed from `"environment" | "file" | "prompt"` to `"env" | "file" | "keychain" | "manual"` to match what `config-validator.ts` accepts
- Updated `isCredentialConfig()` guard, `resolveCredentialFromConfig()` switch cases, `credentialSourceSchema` Zod enum, `getTokenSource()` return type
- `keychain` and `manual` throw explicit not-implemented errors (better than silent fallthrough)
- Added 8 tests covering all four source values

Previously, `source: "env"` would pass validation but silently fall through the resolver's switch. Now the two layers are consistent.

## Test plan
- [x] 8 new tests in `credential-resolver.test.ts` all pass
- [x] TypeScript compiles cleanly
- [x] All pre-commit checks pass (lint, format, security, tests)

### DIFF
--- a/src/domain/configuration/credential-resolver.test.ts
+++ b/src/domain/configuration/credential-resolver.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, test } from "bun:test";
+import { DefaultCredentialResolver } from "./credential-resolver";
+import type { CredentialConfig } from "./credential-resolver";
+
+describe("DefaultCredentialResolver.resolveCredentialFromConfig", () => {
+  const resolver = new DefaultCredentialResolver();
+
+  describe('source: "env"', () => {
+    test("returns token when provided", async () => {
+      const config: CredentialConfig = { source: "env", token: "my-token" };
+      const result = await resolver.resolveCredentialFromConfig(config);
+      expect(result).toBe("my-token");
+    });
+
+    test("returns api_key when token is absent", async () => {
+      const config: CredentialConfig = { source: "env", api_key: "my-api-key" };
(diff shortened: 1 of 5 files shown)


edobry (commented):
## Review: Credential source value alignment (v2)

### Findings

No blocking issues. This is a clean rewrite of the closed PR #357, with the dead-field issue from the first review already fixed (no `env_var`/`path` fields added).

### Checked and clear

- `CredentialSource` type aligned across `credential-resolver.ts`, `types.ts`, and `schemas/base.ts` — all use `["env", "file", "keychain", "manual"]`
- `config-validator.ts` already uses `["env", "file", "keychain", "manual"]` at lines 192 and 277 — confirmed alignment
- `isCredentialConfig` guard updated to match
- `getTokenSource` in `schemas/github.ts` updated from `"environment"` to `"env"`
- `keychain` and `manual` throw explicit not-implemented errors
- 8 tests cover all 4 source values including error paths
- No dead fields in `CredentialConfig` interface — only `token`, `token_file`, `api_key`, `api_key_file`

### Spec verification

**Task:** mt#673

| CredentialSource accepts same values as config-validator | Met | Both use `["env", "file", "keychain", "manual"]` |
| Only "file" overlaps — fix the mismatch | Met | All 4 values now shared |
| Config with source: "env" no longer silently fails | Met | `case "env"` in switch at credential-resolver.ts:86 |
| Add test that catches future drift | Met | 8 tests in `credential-resolver.test.ts` |

PR #383, opened 2026-04-14
feat(mt#713): Add git.log and git.search MCP tools

edobry:
## Summary

- Adds `git.log` command: view commit history with filters for author, date range, file path, message grep, output format (oneline/short/medium/full), and starting ref
- Adds `git.search` command: search repository by content (`git grep`), commit pickaxe (`git log -S`), or diff text (`git log -p --grep`); supports case-insensitive flag and result limit
- Both commands follow the existing pattern in `src/adapters/shared/commands/git.ts` using `composeParams`, `execAsync`, `CommandCategory.GIT`, and are auto-registered in both MCP and CLI adapters

## Test plan

- [x] `bun run validate-all` passes (1515 tests, 0 failures, 0 lint errors)
- [ ] Verify `git.log` and `git.search` MCP tools appear in tool listing
- [ ] Test `git.log` with no params returns last 20 commits in oneline format
- [ ] Test `git.log` with `author`, `since`, `path`, `grep` filters
- [ ] Test `git.search` with `type: "content"` finds pattern in working tree
- [ ] Test `git.search` with `type: "commits"` returns commits touching pattern
- [ ] Test `git.search` with `ignoreCase: true`

### DIFF
--- a/src/adapters/shared/commands/git.ts
+++ b/src/adapters/shared/commands/git.ts
@@ -594,36 +594,13 @@ export function registerGitCommands(): void {
         let command: string;
 
         if (type === "content") {
-          // Use git grep to search working tree / specific ref
           const args: string[] = ["git", "-C", repoPath, "grep"];
           if (ignoreCase) args.push("-i");
-          args.push("-n"); // show line numbers
-          args.push("--", pattern);
-          if (params!.ref) {
-            // git grep searches working tree by default; to search at a ref, pass ref before --
-            // Rebuild: git grep [opts] <pattern> <ref> -- [path]
-            const grepArgs: string[] = ["git", "-C", repoPath, "grep"];
-            if (ignoreCase) grepArgs.push("-i");
-            grepArgs.push("-n");
-            grepArgs.push(pattern);
-            grepArgs.push(params!.ref);
-            if (params!.path) {
-              grepArgs.push("--", params!.path);
-            }
-            command = grepArgs.join(" ");
-          } else {
-            if (params!.path) args.push("--", params!.path);
-            // Remove the duplicate '--' already added before pattern
-            // Rebuild cleanly for working tree search
-            const cleanArgs: string[] = ["git", "-C", repoPath, "grep"];
-            if (ignoreCase) cleanArgs.push("-i");
-            cleanArgs.push("-n");
-            cleanArgs.push(pattern);
-            if (params!.path) {
-              cleanArgs.push("--", params!.path);
-            }
-            command = cleanArgs.join(" ");
-          }
+          args.push("-n");
+          args.push(pattern);
+          if (params!.ref) args.push(params!.ref);
+          if (params!.path) args.push("--", params!.path);
+          command = args.join(" ");
         } else if (type === "commits") {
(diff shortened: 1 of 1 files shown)


edobry (commented):
## Review: git.log and git.search MCP tools

**CI status:** All checks passing (build + Prevent Placeholder Tests)

### Findings

**[BLOCKING]** `git.search` content branch — dead code and rebuild confusion (`src/adapters/shared/commands/git.ts:~580-610`)

The `content` type branch builds an `args` array with `args.push("--", pattern)`, then immediately discards it and rebuilds as either `grepArgs` (ref case) or `cleanArgs` (no-ref case). The initial `args` array is dead code. This should be a single clean construction:

```ts
if (type === "content") {
  const args = ["git", "-C", repoPath, "grep"];
  if (ignoreCase) args.push("-i");
  args.push("-n");
  args.push(pattern);
  if (params!.ref) args.push(params!.ref);
  if (params!.path) args.push("--", params!.path);
  command = args.join(" ");
}
```

**[BLOCKING]** `git.search` diff type uses `--grep` which searches commit *messages*, not diffs (`src/adapters/shared/commands/git.ts:~640`)

The description says "diff (git log -p with grep)" but `--grep=<pattern>` filters by commit message text. To actually search diff content, use `-G<pattern>` (regex) or `-S<pattern>` (pickaxe). The `-p` flag shows the full diff but doesn't filter by it. Either:
- Change to `-G<pattern>` to match the "search in diffs" semantics, or
- Rename the type to clarify it searches commit messages and shows diffs

**[NON-BLOCKING]** `git grep` output has no size limit (`src/adapters/shared/commands/git.ts:~580`)

The `limit` parameter isn't applied to `git grep` (content search). A broad pattern in a large repo could return massive output. Consider piping through `head -n` or using `git grep -c` for count mode, or documenting that `limit` only applies to commit-based searches.

**[NON-BLOCKING]** `git.log` params use `params!.author` etc. without type narrowing — consistent with existing pattern in this file but worth noting for future cleanup.

### Checked and clear

- Parameter definitions follow established `composeParams` pattern correctly
- `CommandCategory.GIT` ensures auto-registration in both MCP and CLI — no wiring changes needed
- `execAsync` import and usage matches existing patterns in git-core-operations.ts
- Error handling for `git grep` exit code 1 (no matches) is correct
- Shell command construction pattern is consistent with pre-existing code in git-core-operations.ts (string interpolation into `exec`)

### Spec verification

**Task:** mt#713

|---|---|---|
| Search commit messages (git log --grep) | Met | `git.log` grep param + `git.search` commits type |
| Filter by author, date range, file paths | Met | `git.log` author/since/until/path params |
| Show commit history (oneline, detailed) | Met | `git.log` format param (oneline/short/medium/full) |
| Diff between refs (git diff ref1..ref2) | Not met | No `git.diff` command implemented |
| Blame (git blame) | Not met | No `git.blame` command implemented |

**Action required:** Two spec criteria are not met (diff between refs, blame). The PR delivers the core value (log + search) and these can be follow-ups. Update the spec to reflect reduced scope and create follow-up task(s) for the remaining capabilities before merging.

PR #384, opened 2026-04-14
feat(mt#637): Add dependsOn parameter to tasks_create for inline dependency declaration

edobry:
## Summary

- Adds optional `dependsOn` parameter to `tasks_create` accepting a single task ID or array of task IDs
- After the task is created, calls `TaskGraphService.addDependency()` for each dependency
- Dependency creation is best-effort: if the persistence provider is unavailable or a dep fails, a warning is included in the output but task creation still succeeds
- `createTasksCreateCommand` now accepts an optional `getPersistenceProvider` factory; `tasks-modular.ts` passes it during registration

## Changes

- `src/adapters/shared/commands/tasks/task-parameters.ts` — added `dependsOn` to `taskCreationParams`
- `src/schemas/tasks.ts` — added `dependsOn` to `taskCreateParamsSchema`
- `src/adapters/shared/commands/tasks/crud-commands.ts` — added `dependsOn` to `TasksCreateParams`, wired dependency creation logic into `TasksCreateCommand.execute()`, updated factory signature
- `src/adapters/shared/commands/tasks-modular.ts` — moved `getPersistenceProvider` declaration before command constructors, passes it to `createTasksCreateCommand`

## Test plan

- [ ] All 1515 existing tests pass (`bun run validate-all` green)
- [ ] `minsky tasks create --title "..." --description "..." --depends-on mt#123` creates task and adds dependency edge
- [ ] `minsky tasks create ... --depends-on mt#123 --depends-on mt#456` (or array form) adds multiple edges
- [ ] If persistence is unavailable, task is still created and a warning is shown

### DIFF
--- a/src/adapters/shared/commands/tasks-modular.ts
+++ b/src/adapters/shared/commands/tasks-modular.ts
@@ -49,9 +49,10 @@ export class ModularTasksCommandManager {
 
       // Create command instances to get their parameter definitions
       log.debug("[ModularTasksCommandManager] Creating command instances");
+      const getPersistenceProvider = () => PersistenceService.getProvider();
       const listCommand = createTasksListCommand();
       const getCommand = createTasksGetCommand();
-      const createCommand = createTasksCreateCommand();
+      const createCommand = createTasksCreateCommand(getPersistenceProvider);
       const editCommand = createTasksEditCommand();
       const deleteCommand = createTasksDeleteCommand();
       const specCommand = createTasksSpecCommand();
@@ -63,8 +64,6 @@ export class ModularTasksCommandManager {
       const similarCommand = new TasksSimilarCommand();
       const searchCommand = new TasksSearchCommand();
       const indexEmbeddingsCommand = new TasksIndexEmbeddingsCommand();
-
-      const getPersistenceProvider = () => PersistenceService.getProvider();
       const depsAddCommand = createTasksDepsAddCommand(getPersistenceProvider);
       const depsRmCommand = createTasksDepsRmCommand(getPersistenceProvider);
       const depsListCommand = createTasksDepsListCommand(getPersistenceProvider);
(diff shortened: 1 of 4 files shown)


edobry (commented):
## Review: dependsOn parameter for tasks_create

**CI status:** All checks passing (build + Prevent Placeholder Tests)

### Findings

**[NON-BLOCKING]** `crud-commands.ts:228` — `const db: PostgresJsDatabase = await persistence.getDatabaseConnection?.();`

The `?.()` can return `undefined` when `getDatabaseConnection` is not defined on the provider, but `db` is typed as `PostgresJsDatabase` (non-nullable). At runtime this is fine — `new TaskGraphService(undefined)` would throw and the outer try/catch handles it gracefully. But the type annotation is slightly misleading. Could be `const db = await persistence.getDatabaseConnection?.();` and let inference handle it, or add an explicit null check before constructing the service.

Minor — the runtime behavior is correct either way.

### Checked and clear

- Parameter schema correctly uses `z.union([z.string(), z.array(z.string())]).optional()` in both task-parameters.ts and schemas/tasks.ts
- Best-effort pattern is well implemented: task creation succeeds even if dep linking fails, with clear warnings
- `getPersistenceProvider` is properly optional — constructor, factory function, and the execute logic all handle its absence
- `tasks-modular.ts` change is minimal and correct — just moves declaration up
- Output includes `depsAdded` and `depsWarnings` in both human-readable and JSON formats
- Error messages use `getErrorMessage()` helper consistently

### Spec verification

**Task:** mt#637

|---|---|---|
| Optional dependsOn parameter (string or string[]) | Met | Added to task-parameters.ts, schemas/tasks.ts, TasksCreateParams interface |
| Calls deps_add internally after creation | Met | Uses TaskGraphService.addDependency() in execute(), lines 227-249 |
| Single atomic operation | Met | One command call creates task + links deps |
| Best-effort if persistence unavailable | Met | Three-tier fallback: no provider → warning, provider error → warning, individual dep error → warning per dep |

All spec criteria met. Clean to merge after the non-blocking type annotation is addressed (or accepted as-is).

PR #410, opened 2026-04-14
feat(mt#727): reference clone optimization for fast session creation

edobry:
## Summary

Adds `git clone --reference` support to session creation, using a local workspace's git objects to avoid re-downloading from the remote. This reduces the clone phase from ~4s to ~0.75s (5x faster on local network; much more on slower connections).

### Environment-aware detection

The MCP server can run in different environments (dev machine, remote server for Claude AI, CI). Detection is layered:

1. **Config-driven** (primary): reads `workspace.mainPath` from `.minsky/config.local.yaml` (gitignored, per-machine)
2. **Heuristic fallback**: tries `process.cwd()` if config not set
3. **Validates**: compares normalized repo names via `normalizeRepositoryUri` (handles SSH vs HTTPS, `.git` suffix)
4. **Graceful degradation**: silently falls back to full clone if no local workspace found or validation fails

### Changes

- **`CloneOptions`** (`types.ts` + `clone-operations.ts`): added `referenceRepo?: string` field; clone command uses `--reference` flag when set
- **`start-session-operations.ts`**: auto-detects local workspace from config or cwd, validates repo identity, passes `referenceRepo` to clone
- **`init.ts` + `config-content.ts`**: `minsky init` now writes `workspace.mainPath` to `.minsky/config.local.yaml` so the optimization works automatically

### Benchmarks

| Method | Time |
|--------|------|
| Full clone (current) | 4.0s |
| Reference clone | 0.75s |
| bun install (warm cache) | 3.4s |
| **Total current** | ~7.4s |
| **Total with reference** | ~4.1s |

## Test plan

- [x] Typecheck passes (`tsc --noEmit`)
- [x] 309 git/session domain tests pass, 0 failures
- [x] Lint: 0 errors (21 pre-existing warnings)
- [ ] Manual test: `minsky session start` with `workspace.mainPath` set - verify `--reference` in clone command (visible in debug logs)
- [ ] Manual test: `minsky session start` without config - verify cwd fallback works
- [ ] Manual test: remote environment without local repo - verify graceful fallback to full clone

### DIFF
--- a/src/domain/git/clone-operations.ts
+++ b/src/domain/git/clone-operations.ts
@@ -10,6 +10,8 @@ export interface CloneOptions {
   workdir: string; // Explicit path where to clone, provided by caller
   session?: string;
   branch?: string;
+  /** Local repo path to use as git --reference source for faster cloning */
+  referenceRepo?: string;
 }
 
 /**
@@ -61,8 +63,9 @@ export async function cloneImpl(
     }
 
     // Clone the repository with verbose logging FIRST
-    log.debug(`Executing: git clone ${options.repoUrl} ${workdir}`);
-    const cloneCmd = `git clone ${options.repoUrl} ${workdir}`;
+    const refFlag = options.referenceRepo ? `--reference "${options.referenceRepo}" ` : "";
+    const cloneCmd = `git clone ${refFlag}${options.repoUrl} ${workdir}`;
+    log.debug(`Executing: ${cloneCmd}`);
 
     // Ensure parent directory exists (handle undefined workdir)
     if (!workdir) {
--- a/src/domain/git/types.ts
+++ b/src/domain/git/types.ts
@@ -209,6 +209,8 @@ export interface CloneOptions {
   workdir: string; // Explicit path where to clone, provided by caller
   session?: string;
   branch?: string;
+  /** Local repo path to use as git --reference source for faster cloning */
+  referenceRepo?: string;
 }
 
 export interface CloneResult {
--- a/src/domain/init.ts
+++ b/src/domain/init.ts
@@ -4,7 +4,11 @@ import { enumSchemas } from "./configuration/schemas/base";
 import { createDirectoryIfNotExists, createFileIfNotExists } from "./init/file-system";
 import type { FsLike } from "./interfaces/fs-like";
 import { createRealFs } from "./interfaces/real-fs";
-import { getMinskyConfigContentYaml, getMCPConfigContent } from "./init/config-content";
+import {
+  getMinskyConfigContentYaml,
+  getLocalConfigContentYaml,
+  getMCPConfigContent,
+} from "./init/config-content";
 import {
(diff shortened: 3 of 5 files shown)


edobry (commented):
## Review: Reference clone optimization for session_start

**CI status:** `Prevent Placeholder Tests` passed, `build` in progress

### Findings

**[PRE-EXISTING]** `src/domain/git/clone-operations.ts:65` — Shell command construction via string interpolation. `repoUrl` and `workdir` are already interpolated without shell escaping. The new `referenceRepo` actually improves on this by quoting the value (`--reference "${options.referenceRepo}"`), but all three values share the same pre-existing pattern. Not introduced by this PR.

### Checked and clear

- **Type consistency**: `CloneOptions.referenceRepo` added in both `types.ts` (canonical) and `clone-operations.ts` (local duplicate) — consistent
- **Config loading**: Verified `config.local.yaml` is loaded by the project config source (`src/domain/configuration/sources/project.ts:20`) as a gitignored overlay — the new `workspace.mainPath` value will be picked up automatically
- **`execInRepository`**: Confirmed it exists on `GitServiceInterface` (`types.ts:45`) — the detection call is valid
- **`normalizeRepositoryUri`**: Confirmed it accepts `UriOptions` parameter and normalizes SSH/HTTPS to `owner/repo` format — comparison logic is correct
- **Graceful degradation**: The entire detection block is wrapped in try/catch with empty catch — any failure (no git, no remote, config missing, parse error) silently falls back to full clone
- **`--repo` interaction**: When `--repo` is passed, `repo` is truthy so the detection block is skipped entirely — no conflict with existing fast-clone path
- **Init**: `createFileIfNotExists` with `overwrite` param means re-running `minsky init` without `--overwrite` won't clobber user's local config — correct behavior
- **Session records**: `repoUrl` (canonical) is still what gets stored on the session record (`start-session-operations.ts:244`) — `referenceRepo` only affects the clone command

### Spec verification

**Task:** mt#727

| Prior clone optimization work identified | Met (pre-PR) | Spec updated with investigation findings |
| Backend detection bug understood and documented | Met (pre-PR) | Spec updated with investigation findings |
| Benchmark of reference clone + warm bun install time | Met (pre-PR) | Spec updated: 0.75s + 3.4s = ~4.1s |
| `workspace.mainPath` wired up: read from config in session_start, written by `minsky init` | Met | `start-session-operations.ts:133` reads config, `init.ts:168` writes `config.local.yaml` |
| `CloneOptions` extended with `referenceRepo`, clone uses `--reference` | Met | `types.ts:212`, `clone-operations.ts:12`, `clone-operations.ts:66` |
| Auto-detection logic: config-driven with cwd fallback, normalized URI comparison | Met | `start-session-operations.ts:128-155` — config first, `currentDir` fallback, `normalizeRepositoryUri` comparison |
| session_start time reduced (clone phase <2s) | Met | Benchmark: 0.75s reference clone vs 4.0s full clone |

All spec criteria met.

PR #414, opened 2026-04-14
fix(mt#735): Surface similarity search backend in output and fix silent error swallowing

edobry:
## Summary

- Replace bare catch in SimilaritySearchService.search() with log.warn() calls reporting which backend failed and why
- Log a warning when all backends fail and empty results are returned
- Change TaskSimilarityService.searchByText() return type to include searchBackend name
- TasksSearchCommand prints "Search backend: lexical (embedding unavailable)" or "Search backend: embeddings" to stderr in text mode, and includes searchBackend in JSON output
- Update searchSimilarTasks() to destructure the new return type
- Update test to use new return shape and assert searchBackend value

## Test plan

- [ ] Run bun test — all 1512 tests pass
- [ ] Run bun run validate-all — typecheck and lint pass
- [ ] Manual: run minsky tasks search without OpenAI key — shows "Search backend: lexical (embedding unavailable)"
- [ ] Manual: run with --json — output includes searchBackend field

### DIFF
--- a/src/adapters/shared/commands/tasks/similarity-commands.ts
+++ b/src/adapters/shared/commands/tasks/similarity-commands.ts
@@ -264,7 +264,30 @@ export class TasksSearchCommand extends BaseTaskCommand<TasksSearchParams> {
       filters.statusExclude = [TaskStatus.DONE, TaskStatus.CLOSED];
     }
 
-    const searchResults = await service.searchByText(query, limit, threshold, filters);
+    const { results: searchResults, searchBackend } = await service.searchByText(
+      query,
+      limit,
+      threshold,
+      filters
+    );
+
+    // Show backend info to stderr unless JSON/quiet
+    try {
+      const { log } = await import("../../../../utils/logger");
+      const quiet = Boolean(params.quiet);
+      const json = Boolean(params.json) || ctx.format === "json";
+      if (!quiet && !json) {
+        const backendLabel =
+          searchBackend === "embeddings"
+            ? "embeddings"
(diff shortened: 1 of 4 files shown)


edobry (commented):
## Review: Search backend observability

**CI status:** In progress

### Findings

No blocking issues.

**[NON-BLOCKING]** `similarity-commands.ts:~274` — Dynamic `await import("../../../../utils/logger")` wrapped in try/catch. The logger is a static module used everywhere — a top-level import would be simpler and consistent with the rest of the codebase. Minor.

### Checked and clear

- `similarity-search-service.ts`: bare `catch` replaced with `catch (error)` + `log.warn` including backend name and error message. Fallthrough also logged.
- `task-similarity-service.ts`: `searchByText()` return type changed to `{ results, searchBackend }`. `searchSimilarTasks()` correctly destructures `{ results }`.
- `similarity-commands.ts`: backend name added to JSON output (`searchBackend` field) and text output (stderr log). Correct conditional for quiet/json modes.
- Test updated to match new return type and assert `searchBackend === "lexical"`.

### Spec verification

**Task:** mt#735

|---|---|---|
| Replace silent `catch { continue }` with logging | Met | `similarity-search-service.ts` — `log.warn` with backend name and error |
| Surface which backend was used in output | Met | JSON: `searchBackend` field. Text: `log.cliWarn` with backend label |
| Log when all backends fail | Met | `log.warn("All similarity search backends failed")` |

All criteria met. Clean to merge.

PR #417, opened 2026-04-14
fix(mt#727): add missing git prefix to execInRepository call in reference clone detection

edobry:
## Summary

Fixes the reference clone optimization from #410. The `execInRepository` method runs raw shell commands, so the command needs the `git` prefix. Without it, `remote get-url origin` fails as a shell command (no such binary), gets caught by the try/catch, and the reference clone optimization is silently skipped.

**Before:** `execInRepository(path, "remote get-url origin")` - fails silently
**After:** `execInRepository(path, "git remote get-url origin")` - works correctly

Confirmed by checking production callers which all use the `git` prefix (e.g., `"git log -1 ..."`, `"git rev-parse HEAD"`, `"git fetch origin"`).

## Test plan

- [x] 309 git/session tests pass
- [ ] Manual test: start a session and verify `.git/objects/info/alternates` exists (proof of --reference clone)

### DIFF
--- a/src/domain/session/start-session-operations.ts
+++ b/src/domain/session/start-session-operations.ts
@@ -136,7 +136,7 @@ Need help? Run 'minsky sessions list' to see all available sessions.`);
         candidatePath = cfg.workspace?.mainPath || currentDir;
 
         const localRemote = (
-          await deps.gitService.execInRepository(candidatePath, "remote get-url origin")
+          await deps.gitService.execInRepository(candidatePath, "git remote get-url origin")
         ).trim();
         if (localRemote) {
           const { normalizeRepositoryUri } = await import("../uri-utils");


edobry (commented):
## Review: Fix missing git prefix in reference clone detection

**CI status:** pending

### Findings

No issues. One-line fix — verified that all production callers of `execInRepository` use the `git` prefix:
- `session-commands.ts:232`: `"git log -1 --pretty=format:..."`
- `session-approve-operations.ts:276`: `"git rev-parse HEAD"`
- `session-review-operations.ts:309`: `"git fetch origin"`

The command `"remote get-url origin"` without `git` fails as `remote: command not found`, which was silently caught by the try/catch and caused the optimization to never activate.

### Checked and clear

- Only one line changed: `"remote get-url origin"` → `"git remote get-url origin"`
- No behavioral change to any other code path
- 309 tests pass

### Spec verification

**Task:** mt#727

| Auto-detection logic: config-driven with cwd fallback, normalized URI comparison | Met (with this fix) | `start-session-operations.ts:139` now correctly calls `git remote get-url origin` |
| All other criteria | Met | Verified in PR #410 review |

PR #411, opened 2026-04-14
chore(mt#690): reach zero ESLint warnings — fix glob matching bug and extract magic strings

edobry:
## Summary

Reaches **zero ESLint warnings** across the entire codebase by fixing three issues:

1. **Glob matching bug in `no-singleton-reach-in` rule**: The hand-rolled `globToRegex()` function failed to match `**` as zero path segments — `**/commands/**/*.ts` didn't match `commands/persistence.ts`. Replaced with `minimatch` library (with `dot: true` for absolute paths).

2. **Missing allowlist entries**: Added 9 legitimate composition-root files that call `getSharedSessionProvider()` to the rule's allowlist.

3. **Magic string duplications**: Extracted `"stash-list stashEntries"` and `"git add process/tasks.md"` to constants in 2 test files.

`MAX_LINT_WARNINGS` set to 0 — zero warnings enforced.

## Test plan

- [x] `tsc --noEmit` — zero errors
- [x] `eslint .` — zero errors, **zero warnings**
- [x] `bun test` — 1512 pass, 0 fail
- [x] Pre-commit hook passes with threshold 0

(Had Claude look into this — AI-assisted cleanup)

PR #420, opened 2026-04-14
feat(mt#739): Add post-merge hook to auto-install when bun.lock changes

edobry:
## Summary

- Adds `.husky/post-merge` hook that detects when `bun.lock` changes after a merge
- Automatically runs `bun install` to keep local packages in sync with the lockfile
- Prevents crashes (like the ESLint failure after PR #411 added `minimatch`) caused by missing packages after pulling merges

## Test plan

- [ ] Merge a PR that adds/updates a dependency and verify `bun install` runs automatically
- [ ] Merge a PR with no lockfile changes and verify `bun install` does NOT run

### DIFF
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,4 +1,4 @@
-if git diff --name-only HEAD@{1} HEAD | grep -q "bun.lock"; then
+if git diff --name-only ORIG_HEAD HEAD | grep -q "bun.lock"; then
   echo "bun.lock changed, running bun install..."
   bun install
 fi
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,0 +1,4 @@
+if git diff --name-only HEAD@{1} HEAD | grep -q "bun.lock"; then
+  echo "bun.lock changed, running bun install..."
+  bun install
+fi


edobry (commented):
## Review: post-merge hook for auto bun install

**CI status:** All checks passed (build ✓, Prevent Placeholder Tests ✓)

### Findings

**[BLOCKING]** `.husky/post-merge` line 1 — `HEAD@{1}` should be `ORIG_HEAD`

In a `post-merge` hook, git guarantees that `ORIG_HEAD` is set to the pre-merge commit. `HEAD@{1}` reads the reflog instead, which works in the common case but fails in at least two real scenarios:

1. **Fresh clone + first pull**: The reflog may only have one entry (`HEAD@{0}`). `git diff --name-only HEAD@{1} HEAD` exits with a fatal error; the pipe swallows it and `grep -q` never sees the output, so `bun install` silently doesn't run even when `bun.lock` changed.
2. **Reflog manipulation** (e.g., after a `git reflog expire` or in CI environments that shallow-clone with `--no-single-branch`): `HEAD@{1}` may point to something unrelated to the merge.

The fix is one word:

```sh
if git diff --name-only ORIG_HEAD HEAD | grep -q "bun.lock"; then
```

`ORIG_HEAD` is explicitly defined by git's merge machinery and is always correct inside `post-merge`. This is the idiomatic pattern shown in git-scm.com docs for post-merge hooks.

### Checked and clear

- **Executable bit**: File is committed as mode `100755` — correct for a git hook.
- **Detection logic**: Diffing the merge's pre- and post-states and grepping for `bun.lock` is the right approach. Both additions and removals of packages change `bun.lock`, so `bun install` is correctly triggered in all dependency-change cases.
- **POSIX compatibility**: The `if/grep -q/fi` pattern is portable sh, consistent with the project's other hooks.
- **No spurious runs**: The `grep -q` short-circuits correctly; `bun install` only runs when `bun.lock` is in the changed-file list.
- **No shebang**: Consistent with other hooks in the repo (`.husky/pre-push` also lacks one), so Husky supplies the shell.

**[NON-BLOCKING]** Style mismatch — the existing hooks use emoji-prefixed echo messages and comment headers (e.g., `# Pre-commit validation hook`). This hook's two-line echo is functional but inconsistent. Worth aligning in a follow-up if style standardization is a concern.

### Spec verification

**Task:** mt#739

|---|---|---|
| Add a post-merge git hook (or Husky hook) | Met | `.husky/post-merge` added, committed as 100755 |
| Detects `bun.lock` changes | Met | `git diff --name-only ... \| grep -q "bun.lock"` |
| Runs `bun install` automatically when changed | Met | `bun install` called inside the `if` block |
| Does NOT run when lockfile unchanged | Met | Gated by `grep -q` — only fires on match |

The spec criterion is functionally delivered, but the `HEAD@{1}` → `ORIG_HEAD` fix is required to make it reliable in edge cases (fresh clone, CI shallow clones).

**Action required:** Fix `HEAD@{1}` → `ORIG_HEAD` before merge.

---

PR #428, opened 2026-04-14
fix(mt#758): Preserve tags during backend migration

edobry:
## Summary

- Adds `tags: fullTask.tags` to the `createTaskFromTitleAndSpec` options call in the migrate-backend command
- Previously, tags were not passed during migration, causing them to be silently dropped when moving tasks between backends

## Test plan

- [x] `bun run validate-all` passes (1474 tests, 0 failures, typecheck clean)
- [ ] Manual: migrate a tagged task between backends and verify tags are preserved in the target backend

### DIFF
--- a/src/adapters/shared/commands/tasks/migrate-backend-command.ts
+++ b/src/adapters/shared/commands/tasks/migrate-backend-command.ts
@@ -316,6 +316,7 @@ export class TasksMigrateBackendCommand extends BaseTaskCommand<MigrateBackendPa
             force: true,
             id: newTaskId,
             status: fullTask.status,
+            tags: fullTask.tags,
           });
 
           // Update session task associations if task ID changed


edobry (commented):
## Review: Preserve tags during backend migration

**CI status:** In progress

### Findings

No issues. Single-line addition of `tags: fullTask.tags` to the migration options object.

### Checked and clear

- `fullTask` is fetched via `sourceService.getTask(taskId)` which returns `Task` including `tags?: string[]`
- `CreateTaskOptions` already has `tags?: string[]` — the field was just missing from the options object
- On GitHub target: tags become labels via mt#725's `createIssueFromTitleAndSpec` which appends tags to labels
- On Minsky target: tags stored in JSON column as before

### Spec verification

**Task:** mt#758

|---|---|---|
| Pass `fullTask.tags` in migration CreateTaskOptions | Met | `migrate-backend-command.ts:319` |

Clean to merge.

PR #429, opened 2026-04-14
refactor(mt#757): consolidate zod error formatters, delete zod-error-formatter.ts

edobry:
## Summary

Two parallel `formatZodError` implementations existed: a rich one in `src/utils/zod-error-formatter.ts` (184 lines, 1 consumer) and a simpler one in `src/domain/schemas/validation-utils.ts` (used internally). Point the single consumer (`parameter-mapper.ts`) at the existing domain-layer implementation and delete the duplicate file.

The simpler formatter is adequate — the caller wraps the result in its own `Invalid value for parameter` message, so the extra type/enum/size-specific formatting from the rich version adds minimal value.

## Test plan

- [x] `tsc --noEmit` — no errors
- [x] Full suite — 1474/1474 pass


edobry (commented):
## Review: Consolidate zod error formatters

**CI status:** pending

### Findings

No blocking issues. Clean consolidation — two implementations with the same name, one consumer, one deleted.

### Checked and clear

- `parameter-mapper.ts:278` — calls `formatZodError(error, paramName)`. Both implementations accept `(ZodError, string?)` — signature compatible.
- The simpler `validation-utils.ts` version formats `message at 'path'` per issue. The rich version added type-specific formatting (enum options, type expected/received, size bounds). Since the caller wraps in `Invalid value for parameter '${paramName}': ${message}`, the extra detail is redundant.
- `validation-utils.ts` already re-exported via `domain/schemas/index.ts` — import path from adapter to domain is architecturally correct.
- No other file imported from `zod-error-formatter.ts`.

### Spec verification

**Task:** mt#757

|---|---|---|
| Duplicate formatter consolidated | Met | Import rewired, file deleted |
| tsc clean | Met | Passed |
| Tests pass | Met | 1474/1474 |

PR #430, opened 2026-04-14
test(mt#759): Add regression test for tag preservation during backend migration

edobry:
## Summary

- Adds regression test proving tags are passed through `createTaskFromTitleAndSpec` during backend migration (bug mt#758)
- Adds DI seam (`createTaskServiceFactory`) to `TasksMigrateBackendCommand` so tests can inject fake task services without `mock.module()`
- Verified: test fails when `tags: fullTask.tags` is removed, passes when present

## Test plan

- [x] Test passes with fix in place
- [x] Test fails when fix is removed (verified manually)
- [x] `bun run validate-all` passes (1476 tests, 0 failures)

Generated with Claude Code (had Claude implement this)

### DIFF
--- a/src/adapters/shared/commands/tasks/migrate-backend-command.ts
+++ b/src/adapters/shared/commands/tasks/migrate-backend-command.ts
@@ -228,6 +228,9 @@ export class TasksMigrateBackendCommand extends BaseTaskCommand<MigrateBackendPa
     return detectedBackend as string;
   }
 
+  /** @internal — exposed as non-private for test DI injection */
+  createTaskServiceFactory = createConfiguredTaskService;
+
   private async migrateTasksBetweenBackends(options: {
     sourceBackend: string;
     targetBackend: string;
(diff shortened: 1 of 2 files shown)


edobry (commented):
## Review: Migration tag regression test

### Findings

No issues.

### Checked and clear

- DI seam `createTaskServiceFactory` is a public field defaulting to `createConfiguredTaskService` — tests can replace it without `mock.module()`
- Test injects fake source service (returns task with tags) and fake target service (captures `createTaskFromTitleAndSpec` calls)
- Key assertion: `options?.tags` equals `["di-cleanup", "test-quality"]` — directly proves the fix from mt#758
- Manually verified: test fails when `tags: fullTask.tags` is removed from line 319
- Second test covers the no-tags case (undefined tags don't crash)

### Spec verification

**Task:** mt#759

|---|---|---|
| Test that fails when fix is removed | Met | Verified — `expect(options?.tags).toEqual(...)` fails with `undefined` |
| Test that passes with fix in place | Met | 2 pass, 0 fail |
| Uses DI pattern (no mock.module) | Met | `createTaskServiceFactory` field injection |

PR #434, opened 2026-04-17
feat(mt#761): implement typed DI container (Phase 1)

edobry:
## Summary

Phase 1 of mt#761: Build the typed DI container. No changes to existing code — only new files in `src/composition/`.

### What was built

**`src/composition/types.ts`** — The `AppServices` interface (service map) and container interface. All imports are `import type` — zero runtime cost. Defines 7 service keys mapping to existing domain interfaces (`BasePersistenceProvider`, `SessionProviderInterface`, `GitServiceInterface`, etc.).

**`src/composition/container.ts`** — `AppContainer` class (~80 lines):
- `register(key, factory, options?)` — stores a factory (sync or async) for a service key
- `set(key, instance)` — directly injects an instance (for tests)
- `get(key)` — returns cached instance, throws if not resolved
- `has(key)` — checks if a service is available
- `initialize()` — resolves all registered factories in registration order; factories can call `get()` for previously resolved deps; `set()` values take precedence (factories skipped)
- `close()` — calls registered disposers in reverse order, then clears all instances

**`src/composition/container.test.ts`** — 17 unit tests covering all public methods and behavioral contracts.

**`src/composition/index.ts`** — Barrel export.

### Key design decisions (from mt#761 spec)

1. **DIY not a library**: The codebase has 18+ DI interfaces and is function-oriented. A lightweight container fits better than decorator-based frameworks (tsyringe) or paradigm shifts (Effect).
2. **Async init is first-class**: `initialize()` wraps all factories in `Promise.resolve()` — both sync and async factories work uniformly. This addresses the gap in TypeScript DI libraries where JS constructors can't be async.
3. **Registration order = dependency order**: Explicit, debuggable, no topological sort needed for ~20 services.
4. **Reverse-order disposal**: `close()` tears down leaves before roots (sessionProvider before persistence).
5. **Domain code unchanged**: The container is only used in composition roots. Domain code receives typed deps interfaces.

## Test plan

- [x] 17 container unit tests pass
- [x] 1506 existing tests still pass (verified by pre-push hook)
- [x] ESLint: 0 errors, 0 warnings
- [x] No changes to existing files

### DIFF
--- a/src/composition/container.test.ts
+++ b/src/composition/container.test.ts
@@ -1,0 +1,258 @@
+/**
+ * Unit tests for AppContainer
+ *
+ * Tests cover every public API method and the key behavioral contracts
+ * documented in the mt#761 spec:
+ * - register() + initialize() + get() basic flow
+ * - set() directly sets instances (test pattern)
+ * - get() throws for unresolved services
+ * - initialize() resolves in registration order
+ * - Factories can access previously resolved services via get()
+ * - close() calls disposers in reverse order
+ * - set() takes precedence over register()
+ * - has() returns correct boolean
+ * - Async factories are properly awaited
+ */
+
+import { describe, test, expect } from "bun:test";
+import { AppContainer } from "./container";
+import type { AppServices } from "./types";
+
+// Minimal fakes that satisfy the type constraints
+const fakePersistence = {
+  initialize: async () => {},
+  close: async () => {},
+  getStorage: () => ({}) as any,
+  isInitialized: () => true,
+} as unknown as AppServices["persistence"];
+
+const fakeSessionProvider = {
+  getSession: async () => null,
+  listSessions: async () => [],
+  addSession: async () => {},
+  deleteSession: async () => {},
+} as unknown as AppServices["sessionProvider"];
+
+const SERVICE_NOT_AVAILABLE = 'Service "persistence" is not available';
+
+const fakeGitService = {
+  clone: async () => ({ workdir: "", session: "" }),
+  execInRepository: async () => "",
+} as unknown as AppServices["gitService"];
+
+describe("AppContainer", () => {
+  // --- register() + initialize() + get() ---
+
+  test("register + initialize + get: basic sync factory", async () => {
+    const container = new AppContainer();
+    container.register("persistence", () => fakePersistence);
+    await container.initialize();
+    expect(container.get("persistence")).toBe(fakePersistence);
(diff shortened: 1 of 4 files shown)


edobry (commented):
## Review: Typed DI container (Phase 1)

**CI status:** both checks passed (build + Prevent Placeholder Tests)

### Findings

No blocking issues. This is a pure addition — 4 new files in `src/composition/`, zero changes to existing code.

**[NON-BLOCKING]** `container.ts:62` — `get()` uses `instance !== undefined` check, which means a service explicitly set to `undefined` would cause a false "not available" error. Not a practical concern (no service interface returns undefined), but `has()` uses `instances.has(key)` while `get()` uses value check — the semantics are subtly different if someone calls `set("key", undefined as any)`. Edge case that's unlikely to matter.

### Checked and clear

- **Type imports**: All 7 type imports in `types.ts` verified against actual codebase interfaces (`BasePersistenceProvider` at `persistence/types.ts:57`, `SessionProviderInterface` at `session/types.ts:95`, `TaskServiceInterface` at `tasks/taskService.ts:17`, `GitServiceInterface` at `git/types.ts:22`, `WorkspaceUtilsInterface` at `workspace.ts`, `SessionDeps` at `session-service.ts:51`, `RepositoryBackendType` at `repository/`)
- **All imports are `import type`**: Zero runtime cost, no startup impact
- **Container implementation**: ~80 lines, clean separation of concerns. `register()`, `set()`, `get()`, `has()`, `initialize()`, `close()` all behave as documented
- **Registration order tracking**: Maintained via `registrationOrder` array with dedup on re-registration
- **Reverse-order disposal**: `close()` reverses the array — correct for teardown (leaves before roots)
- **Test coverage**: 17 tests covering all public methods, dependency ordering, error cases, chaining, re-registration, async factories, and disposal order
- **No existing files modified**: Confirmed via `git diff --stat`
- **Barrel export**: `index.ts` exports class + types with proper `export type` for type-only re-exports

### Spec verification

**Task:** mt#761

| P1: AppContainer with typed register/get/initialize/close | Met | `container.ts`: 97 lines, implements `AppContainerInterface` |
| P1: Unit tested | Met | `container.test.ts`: 17 tests, 0 failures |
| P1: No changes to existing code | Met | Only new files in `src/composition/` |
| P2-P4 criteria | N/A | Phase 1 only; P2+ are follow-up work |

PR #436, opened 2026-04-18
feat(mt#238): add parent-child task relationships (subtask support)

edobry:
## Summary

Adds typed edge support to the task relationship graph, enabling parent-child relationships alongside existing dependency edges. This is the prerequisite for mt#687 (eliminating the delete-restart anti-pattern).

### Changes

- **Schema**: Added `type` column to `task_relationships` table (`"depends"` | `"parent"`), 3-column unique constraint, partial unique index enforcing one parent per child
- **TaskGraphService**: New parent-child operations (`addParent`, `removeParent`, `getParent`, `listChildren`, `getAncestors`) with cycle prevention
- **`tasks.create --parent`**: Create subtasks inline, following the `--dependsOn` pattern (mt#637)
- **`tasks.children` / `tasks.parent`**: New query commands for navigating hierarchy
- **`tasks.list --hierarchical`**: Tree view grouping subtasks under parents
- **Backward compatible**: Existing dependency operations filter by `type = 'depends'` by default; all callers updated

### Design decisions (documented in spec)

- D1: Child→parent edge direction (consistent with depends convention)
- D2: 3-column unique constraint (composition ≠ sequencing, both can coexist)
- D3: No dotted IDs — subtasks get regular sequential IDs
- D4: No special branch naming — distinct IDs prevent collisions naturally
- D5: `--parent` on existing create, not a separate command
- D9: Default type `'depends'` preserves backward compatibility
- D11: Cycle prevention for parent-child only (full graph cycle detection is mt#239)

### Also fixes

- Pre-existing type error: added missing `tags` field to `TaskData` interface

## Test plan

- [x] 26 unit tests pass covering dependencies, parent-child, cycle prevention, isolation, bulk queries
- [x] All existing task domain tests (43) and adapter tests (8) pass
- [ ] Run migration on Supabase
- [ ] End-to-end: `tasks create --parent mt#237` → verify with `tasks children mt#237`


edobry (commented):
## Review: Parent-child task relationships (subtask support)

**CI status:** In progress (build + Prevent Placeholder Tests running)

### Findings

**[NON-BLOCKING]** `crud-commands.ts` — D7 (orphaning children on parent delete) is specified but not implemented. `TasksDeleteCommand` doesn't clean up parent edges when a task is deleted. Not blocking because: (1) deleting tasks with children is rare, (2) orphan edges are harmless (getParent returns a dangling ID, no crash), (3) can be a follow-up. Worth a task.

**[NON-BLOCKING]** `crud-commands.ts:166` — The `depthMap` is undefined when `hierarchical` is true but `getPersistenceProvider` is not available (e.g., no SQL backend). The formatting block at line 185 checks `params.hierarchical` without checking `depthMap`, so it would render all tasks at depth 0. This is acceptable degradation but could be confusing — a warning message would help.

### Checked and clear

- **Schema migration** (`0017_task_relationship_types.sql`): Correct — adds column with default, drops/recreates unique index as 3-column, adds partial unique index for one-parent enforcement. Statement breakpoints correct.
- **Drizzle schema** (`task-relationships.ts`): Matches migration. Partial index uses `sql` template correctly.
- **TaskGraphService**: All existing dependency methods correctly scoped to `type = "depends"`. New parent-child methods correctly scoped to `type = "parent"`. Cycle prevention walks ancestors iteratively with max depth guard. Idempotency handled correctly (returns `{ created: false }` for existing edges).
- **Backward compatibility**: `deps-rendering.ts` (both call sites), `task-routing-service.ts` — all updated to pass `"depends"` filter. No callers of bulk operations left unfiltered that would break.
- **Test coverage**: 26 tests covering deps, parent-child, cycle prevention, type isolation, bulk queries. In-memory repo correctly models typed edges.
- **New commands**: `tasks.children` and `tasks.parent` properly registered in `registry-setup.ts`. Simple query commands with clean output formatting.
- **`--parent` on tasks.create**: Follows the `--dependsOn` pattern exactly. Error handling is consistent.

### Spec verification

**Task:** mt#238

|---|---|---|
| `task_relationships` has `type` column | Met | `0017_task_relationship_types.sql` adds column, `task-relationships.ts` schema updated |
| `TaskGraphService` supports `addParent`, `getParent`, `listChildren` with cycle prevention | Met | `task-graph-service.ts:177-230`, cycle check at line 184, 26 tests pass |
| Users can create subtasks via `--parent` on `tasks.create` | Met | `crud-commands.ts:362-386`, parameter added to `task-parameters.ts` |
| `tasks.children` command lists subtasks | Met | `deps-commands.ts:169-190` |
| `tasks list --hierarchical` shows tree structure | Met | `crud-commands.ts:118-175`, `task-parameters.ts` hierarchical param |
| All existing dependency operations unchanged | Met | All deps methods filter by `"depends"`, bulk callers updated |
| End-to-end lifecycle works | Not yet verified | Requires migration on Supabase + manual test |

**Action required:** End-to-end verification needed post-merge (migration + `tasks create --parent` test). This is a runtime verification, not a code gap.

PR #438, opened 2026-04-18
feat(mt#728): add session.generate_prompt MCP tool for subagent dispatch

edobry:
## Summary

- Add `session.generate_prompt` MCP tool that generates complete, correct subagent prompt strings
- Implements 4 prompt types: `implementation`, `refactor`, `review`, `cleanup`
- Each generated prompt includes session dir, session UUID, task ID, commit/PR instructions, and guard rails
- Warns when scope exceeds 40 files (subagent context budget)
- 29 tests covering all prompt types, scope handling, and edge cases

## Motivation

The prompt is the interface contract between orchestrator and subagent. Currently hand-crafted from behavioral rules (memory files). mt#672 proved this breaks under load: wrong task IDs, missing sessionIds, unbounded scope. This tool makes correct prompts the easy path.

See [Notion design note](https://www.notion.so/346937f03cb481908efdf88ea877c125) for the full analysis.

## Test plan

- [x] 29 unit tests for `generateSubagentPrompt()` covering all 4 types
- [ ] Manual test: generate a prompt via MCP, dispatch via Agent tool, verify subagent commits with correct task ID

(Had Claude look into this — AI-assisted implementation)

### DIFF
--- a/src/adapters/shared/commands/session.ts
+++ b/src/adapters/shared/commands/session.ts
@@ -32,6 +32,7 @@ import {
 import { createSessionConflictsCommand } from "./session/conflicts-command";
 import { createSessionRepairCommand } from "./session/repair-command";
 import { createSessionEditFileCommand } from "./session/file-commands";
+import { createSessionGeneratePromptCommand } from "./session/prompt-command";
 import { registerSessionChangesetCommands } from "./session/changeset-aliases";
 import { sharedCommandRegistry, type CommandDefinition } from "../command-registry";
 
@@ -87,6 +88,7 @@ export async function registerSessionCommands(
     // Utility
     createSessionConflictsCommand(getDeps),
     createSessionRepairCommand(getDeps),
+    createSessionGeneratePromptCommand(getDeps),
 
     // File
     createSessionEditFileCommand(getDeps),
--- a/src/adapters/shared/commands/session/prompt-command.ts
+++ b/src/adapters/shared/commands/session/prompt-command.ts
@@ -62,6 +62,7 @@ export function createSessionGeneratePromptCommand(getDeps: LazySessionDeps): Co
               .split(",")
               .map((s) => s.trim())
               .filter((s) => s.length > 0)
+              .map((s) => (s.startsWith("/") ? s : `${sessionDir}/${s}`))
           : undefined;
 
       const taskId = task.replace(/^mt#/, "").replace(/^#/, "");
--- a/src/adapters/shared/commands/session/prompt-command.ts
+++ b/src/adapters/shared/commands/session/prompt-command.ts
@@ -1,0 +1,81 @@
+/**
+ * Session Generate Prompt Command
+ *
+ * Generates complete subagent prompt strings for session work.
+ */
+import { CommandCategory, type CommandDefinition } from "../../command-registry";
+import { type LazySessionDeps, withErrorLogging } from "./types";
+import { z } from "zod";
+
+const promptCommandParams = {
+  task: { schema: z.string(), description: "Task ID (required)", required: true },
+  type: {
+    schema: z.enum(["implementation", "refactor", "review", "cleanup"]),
+    description: "Prompt type: implementation, refactor, review, or cleanup",
+    required: true,
+  },
+  instructions: {
+    schema: z.string(),
+    description: "Specific work instructions for the subagent",
+    required: true,
+  },
+  scope: {
+    schema: z.string(),
+    description: "Comma-separated list of file paths to constrain to",
+    required: false,
+  },
+};
+
+export function createSessionGeneratePromptCommand(getDeps: LazySessionDeps): CommandDefinition {
+  return {
+    id: "session.generate_prompt",
+    category: CommandCategory.SESSION,
+    name: "generate_prompt",
+    description: "Generate a complete subagent prompt for session work",
+    parameters: promptCommandParams,
+    execute: withErrorLogging("session.generate_prompt", async (params) => {
+      const { getSessionFromParams } = await import("../../../../domain/session");
+      const { generateSubagentPrompt } = await import(
+        "../../../../domain/session/prompt-generation"
+      );
+      const { resolveSessionDirectory } = await import(
+        "../../../../domain/session/resolve-session-directory"
+      );
+
+      const task = params.task as string;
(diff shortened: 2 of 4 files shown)


edobry (commented):
## Review: session.generate_prompt MCP tool (mt#728 Phase 1)

**CI status:** `build` check in progress; `Prevent Placeholder Tests` passed. Do not merge until `build` completes.

---

### Findings

**[NON-BLOCKING]** `src/adapters/shared/commands/session/prompt-command.ts:55` — `getDeps` parameter accepted but never used

The factory function signature is `createSessionGeneratePromptCommand(getDeps: LazySessionDeps)` but `getDeps` is never called inside the execute handler. All dependencies are loaded via dynamic `import()` instead. This is inconsistent with the pattern every other session command follows, where `getDeps()` is called to obtain the session provider. For this command it is harmless (the imports bypass the lazy cache correctly), but the unused parameter creates a misleading API surface. Consider either removing the parameter and dropping the factory pattern (if DI isn't needed) or actually threading deps through the same way as sibling commands.

**[NON-BLOCKING]** `src/domain/session/prompt-generation.ts:37` — scope files listed as relative paths, but the header instructs "all paths MUST be absolute"

The `renderScopeSection` function lists scope entries verbatim (e.g., `- src/foo.ts`). The caller supplies them raw from the MCP parameter. The header tells the subagent all paths must be absolute, but the scope list may contain relative paths, creating a contradiction in the generated prompt. There is no validation or normalization. This is a usability issue, not a functional bug (the subagent interprets them), but it can confuse subagents already known to guess paths incorrectly. The spec says scope is "a list of file paths to constrain to" — the tool should either accept only absolute paths (and validate) or auto-prefix with `sessionDir`.

**[NON-BLOCKING]** `src/domain/session/prompt-generation.ts:55–67` — `renderCommitInstructions` uses backtick-wrapped inline parameter style that may not render as code in all contexts

The commit instruction renders as:
```
- Parameters: `sessionId: "abc-123"`, `all: true`
```
When the prompt is passed to an Agent tool invocation (which does not parse Markdown), these backticks appear literally. They do not hurt correctness, but the spec's intent ("explicit `sessionId` and `task` parameters") is met — the values are present and unambiguous.

**[NON-BLOCKING]** `tests/domain/session/prompt-generation.test.ts` — no test for the `taskId` stripping logic in `prompt-command.ts`

The command-layer strips `mt#` / `#` prefixes before calling `generateSubagentPrompt`. This stripping is only tested indirectly through the domain function (which receives a pre-stripped `taskId`). A test that exercises `prompt-command.ts` directly (or documents the expected stripping behavior) would prevent regression if the regex is changed. Since the domain tests are thorough for the pure function, this is low risk.

---

### Checked and clear

- **MCP auto-exposure path**: `createSessionGeneratePromptCommand` returns a `CommandDefinition` with `category: CommandCategory.SESSION`. `registerSessionCommandsWithMcp` (via `shared-command-integration.ts`) iterates all SESSION-category commands and calls `commandMapper.addCommand()`. The `id: "session.generate_prompt"` becomes the MCP tool name as-is. The routing is correct.
- **Command registration**: The import and the push into the `commands[]` array in `session.ts` are correct and in the right location.
- **Required metadata in generated prompts**: `sessionDir`, `sessionId`, `taskId` (prefixed with `mt#`) are all present in the header and in the commit/PR sections. Verified against the template source directly.
- **Guard rails by type**: Review type correctly omits `renderCommitInstructions` and includes explicit "Do NOT make any changes" guard. Verified by reading the `generateSubagentPrompt` branching logic — the review early-return path never reaches `renderCommitInstructions`.
- **4 prompt types**: `implementation`, `refactor`, `review`, `cleanup` — all implemented with distinct guard rails.
- **Scope warning at >40 files**: Threshold is `> SCOPE_WARNING_THRESHOLD` (strict greater-than). Test at exactly 40 files expects `undefined` (correct). Test at 41 files expects defined (correct). Off-by-one is correct.
- **`getSessionFromParams({ task })` lookup**: `SessionGetParams` accepts `task` as optional; the schema refine requires either `name` or `task`. The task-based lookup is a supported first-class path through `SessionService.get()`.
- **Task ID stripping**: `task.replace(/^mt#/, "").replace(/^#/, "")` correctly handles `mt#456` → `456` and `#456` → `456`. Plain `456` is unchanged. The domain function then prepends `mt#` back in template strings, producing consistent output.
- **`resolveSessionDirectory`**: Calls `getSharedSessionProvider()` → `provider.getSession()` → `provider.getRepoPath()`. Returns an absolute filesystem path. The path is passed directly into the prompt as `sessionDir`. Correct.
- **Tests are meaningful**: 29 tests, no placeholder assertions. Each test checks a specific behavioral property of the template. The `describe` structure mirrors the type branching in the implementation.

---

### Spec verification

**Task:** mt#728 — Phase 1

|---|---|---|
| `session.generate_prompt` tool exists and is callable via MCP | Met | `id: "session.generate_prompt"` in `prompt-command.ts`; registered in `session.ts`; `CommandCategory.SESSION` auto-exposes via `shared-command-integration.ts` `registerSessionCommandsWithMcp` |
| Generated prompts include explicit `sessionId` and `task` parameters | Met | `renderCommitInstructions` embeds `sessionId: "${sessionId}"` and `task: "mt#${taskId}"` literals in commit/PR sections; verified in template source |
| Generated prompts include absolute session directory path | Met | `sessionDir` from `resolveSessionDirectory()` is an absolute path; embedded in `renderCommonHeader` as `"working in Minsky session at ${params.sessionDir}"` |
| Generated prompts include appropriate guard rails by type | Met | `review` → no commit/PR, read-only guard; `refactor` → coherence verification; `cleanup` → batching guidance; `implementation` → full workflow; all types include no-Bash guard |
| At least 4 prompt types implemented | Met | `implementation`, `refactor`, `review`, `cleanup` — all four implemented with distinct sections |
| Manual test: generate prompt, dispatch via Agent tool, verify subagent commits with correct task ID | Not met | Spec item unchecked in PR description; this is a manual verification step that has not been completed per the PR's own test plan |

**Action required on unchecked manual test:** The spec requires a live end-to-end test (generate prompt → dispatch → verify subagent commits with correct task ID). The PR notes this as unchecked. This should be completed before marking the task DONE, but it is not a blocking merge concern for Phase 1 code quality.

---

PR #443, opened 2026-04-18
feat(mt#782): Persist branch name in session record at creation time

edobry:
## Summary

- Un-deprecates the `branch` field on `SessionRecord` and moves it from legacy/compatibility to a core field
- Stores the computed branch name in the session record during `startSessionImpl` (previously computed but deliberately not persisted)
- Adds branch persistence to the orphan session self-repair path in `session-update-operations.ts`
- Fixes a broken import of `taskIdToBranchName` in `session-update-operations.ts`

## Motivation

Prerequisite for mt#756 (session_delete should clean up remote branches). Without reliably stored branch names, deletion would have to re-derive the branch from the task ID, which is fragile — explicit `--branch` overrides would be lost, and the derivation logic could change over time.

## Test plan

- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] Existing session start and mutability tests pass (11/11)
- [ ] CI build + tests pass

### DIFF
--- a/src/domain/session/session-start-operations.ts
+++ b/src/domain/session/session-start-operations.ts
@@ -190,21 +190,23 @@ Need help? Run 'minsky sessions list' to see all available sessions.`);
       }
     }
 
-    // Prepare session record but don't add to DB yet (branch no longer persisted)
+    // Define branchName before session record so it can be persisted
+    const branchName = branch || (taskId ? taskIdToBranchName(taskId) : sessionId);
+
+    // Prepare session record but don't add to DB yet
     const baseSessionRecord: SessionRecord = {
       session: sessionId,
       repoUrl,
       repoName,
       createdAt: new Date().toISOString(),
       taskId: taskId ? SessionBackwardCompatibility.toStorageFormat(taskId) : undefined,
+      branch: branchName,
     };
 
     // Enhance session record with multi-backend information
     const sessionRecord = SessionMultiBackendIntegration.enhanceSessionRecord(baseSessionRecord);
 
     let sessionAdded = false;
-    // Define branchName outside try block so it's available in return statement
-    const branchName = branch || (taskId ? taskIdToBranchName(taskId) : sessionId);
 
     try {
       // First clone the repo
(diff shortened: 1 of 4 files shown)


edobry (commented):
## Review: Persist branch name in session record

**CI status:** `Prevent Placeholder Tests` ✅ | `build` ⏳ in progress

### Findings

**[NON-BLOCKING]** `src/domain/session/session-start-operations.ts:193` — Dead code file still has stale "branch no longer persisted" comment and does not store branch. This file is not imported anywhere (confirmed via grep — zero import references). Pre-existing dead code, not introduced by this PR. Consider removing the entire file in a follow-up.

### Checked and clear

- **`types.ts`**: `branch` field correctly moved from deprecated legacy section to core fields with proper JSDoc. No duplicate field remaining.
- **`start-session-operations.ts`**: `branchName` computation moved before session record creation. Still in scope for later usage at line 300 (`branchWithoutSession`). Semantically identical — no behavior change beyond persisting the value.
- **`session-update-operations.ts`**: Orphan repair path now stores `branch`. Import of `taskIdToBranchName` correctly points to `../tasks/task-id` (verified export exists at `src/domain/tasks/task-id.ts:126`). The previous import `from "../session"` was broken (module does not export this symbol — confirmed via grep).
- **`session-auto-repair.ts:210`**: Already stores `branch` — no change needed. Confirmed consistent with this PR's approach.
- **Scope**: All three production paths that create session records now store `branch`. No other `addSession` call sites in production code are missing it.

### Spec verification

**Task:** mt#782

| --- | --- | --- |
| Store branch name in session record at creation time in `start-session-operations.ts` | Met | `start-session-operations.ts:281` — `branch: branchName` added to record |
| Un-deprecate `branch` field in `types.ts` | Met | `types.ts:21` — moved to core fields with JSDoc, removed from legacy section |
| Fix orphan repair path in `session-update-operations.ts` to store branch | Met | `session-update-operations.ts:123` — `branch: taskId ? taskIdToBranchName(taskId) : sessionId` |

All spec criteria met. No blocking issues found.

PR #444, opened 2026-04-18
feat(mt#761): thread DI container through command execution (Phase 3a)

edobry:
## Summary

Phase 3a of mt#761: Thread the DI container through the command execution infrastructure so execute handlers can resolve services from it instead of calling singletons.

### Changes

1. **`CommandExecutionContext`** (`command-registry.ts`): Added optional `container` field so all execute handlers can access DI services via `context.container?.get("serviceName")`.

2. **Container injection for CLI** (`command-generator-core.ts`): Module-scoped `setAppContainer()`/`getAppContainer()` functions. Set once from `cli.ts`, injected into every `CliExecutionContext`.

3. **Container injection for MCP** (`shared-command-integration.ts`): MCP context creation uses `getAppContainer()` to include the container.

4. **`registerAllSharedCommands(container?)`** (`commands/index.ts`): Accepts optional container, threads it to `registerSessionCommands()`.

5. **Session `LazySessionDeps`** (`session.ts`): Now resolves from `container.get("sessionDeps")` when available, falling back to `getSharedSessionProvider()` singleton during migration.

6. **Changeset aliases** (`changeset-aliases.ts`): 2 `getSharedSessionProvider()` calls replaced with `ctx?.container?.get("sessionProvider")` with singleton fallback.

### Migration pattern established

Execute handlers can now access the container via their existing `context` parameter:
```typescript
execute: async (params, context) => {
  const sessionProvider = context.container?.has("sessionProvider")
    ? context.container.get("sessionProvider")
    : await getSharedSessionProvider(); // fallback during migration
}
```

This pattern will be applied to remaining singleton call sites in subsequent batches.

## Test plan

- [x] 1555 tests pass (pre-push verified)
- [x] CLI functional test: session --help (0.23s), tasks list works
- [x] No startup regression

### DIFF
--- a/src/adapters/mcp/shared-command-integration.ts
+++ b/src/adapters/mcp/shared-command-integration.ts
@@ -151,10 +151,21 @@ export function registerSharedCommandsWithMcp(
 
           try {
             // Create execution context for shared command
+            // Get the app container set by the composition root (mt#761).
+            let appContainer;
+            try {
+              const { getAppContainer } = await import(
+                "../shared/bridges/cli/command-generator-core"
+              );
+              appContainer = getAppContainer();
+            } catch {
+              // Container not available — backward compatibility
+            }
             const context: CommandExecutionContext = {
               interface: "mcp",
               debug: Boolean(args?.debug),
               format: args?.json === "true" ? "json" : "text", // Use json format only when explicitly requested
+              container: appContainer,
             };
             log.debug(`[MCP] Created execution context: ${command.id}`, { context });
 
--- a/src/adapters/shared/bridges/cli/command-generator-core.ts
+++ b/src/adapters/shared/bridges/cli/command-generator-core.ts
@@ -42,6 +42,27 @@ export interface CommandGeneratorDependencies {
 }
 
 /**
+ * Module-scoped container reference, set once from the CLI composition root.
+ * Injected into CommandExecutionContext so execute handlers can access DI services.
+ * @see mt#761
+ */
+let _appContainer: import("../../../../composition/types").AppContainerInterface | undefined;
+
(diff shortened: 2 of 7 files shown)


edobry (commented):
## Review: Thread DI container through command execution (Phase 3a)

**CI status:** both checks passed

### Findings

No blocking issues.

**[NON-BLOCKING]** `command-generator-core.ts` — The `setAppContainer`/`getAppContainer` module-scoped pattern is a controlled singleton (set once from composition root, read by context creators). It's pragmatic but should be documented as transitional — ideally the container would flow through constructor injection of `CommandGeneratorCore`. Acceptable for now.

### Checked and clear

- **`CommandExecutionContext.container`**: Optional field, backward compatible — existing handlers that don't use it are unaffected
- **CLI context injection**: `setAppContainer()` called in `createCli()` before any commands execute. The `_appContainer` reference is available by the time `preAction` hook initializes the container.
- **MCP context injection**: Uses `getAppContainer()` with try/catch fallback — won't break if the container isn't available
- **Session LazySessionDeps**: `container.has("sessionDeps")` check before `container.get()` — safe, falls back to singleton
- **Changeset aliases**: Both call sites use `ctx?.container?.has("sessionProvider")` pattern with singleton fallback — safe, backward compatible
- **Startup time**: 0.23s session --help — no regression
- **Tests**: 1555 pass, 0 failures

### Spec verification

**Task:** mt#761

| P3a: session command singleton calls migrated | Partially met | `session.ts:55` resolved via container; `changeset-aliases.ts:97,205` use ctx.container with fallback. Full elimination requires removing fallback paths once all contexts set the container. |
| Container accessible to execute handlers | Met | `context.container` available in both CLI and MCP execution contexts |

PR #445, opened 2026-04-18
fix(mt#784): use .returning() for DELETE count instead of broken rowCount

edobry:
## Summary

Drizzle ORM's DELETE result strips postgres.js metadata (`count`, `command`), so `(res as { rowCount }).rowCount` is always `undefined`, falling back to 0. This caused `removeDependency` and `removeParent` to always return `{ removed: false }` even when the delete succeeded.

### Root cause

- **Raw postgres.js** DELETE returns `{ count: 1, command: "DELETE" }` ✅
- **Drizzle ORM** DELETE returns `[]` with no `rowCount` or `count` ❌

### Fix

Use `.returning({ id: taskRelationshipsTable.id })` on DELETE queries and check `deleted.length`. This is the idiomatic Drizzle approach that works through connection poolers like Supabase.

### Changes

- `deleteEdge`: `.returning()` + `.length` instead of `rowCount`
- `deleteEdgesFrom`: same pattern

## Spec verification

|---|---|---|
| DELETE operations return correct count | Met | `.returning()` returns actual deleted rows |
| Backward compatible | Met | Only the count mechanism changed, not behavior |

### DIFF
--- a/src/domain/storage/backends/postgres-storage.ts
+++ b/src/domain/storage/backends/postgres-storage.ts
@@ -398,11 +398,10 @@ export class PostgresStorage implements DatabaseStorage<SessionRecord, SessionDb
    */
   async deleteEntity(id: string): Promise<boolean> {
     await this.ensureConnection();
-    const result = await this.drizzle!.delete(postgresSessions).where(
-      eq(postgresSessions.session, id)
-    );
-    const rowCount = (result as { rowCount?: number }).rowCount ?? 0;
-    return rowCount > 0;
+    const deleted = await this.drizzle!.delete(postgresSessions)
+      .where(eq(postgresSessions.session, id))
+      .returning({ session: postgresSessions.session });
+    return deleted.length > 0;
   }
 
   /**
--- a/src/domain/tasks/github-issues-api.ts
+++ b/src/domain/tasks/github-issues-api.ts
@@ -366,9 +366,12 @@ export async function deleteIssue(
         if (db) {
           const { tasksTable } = await import("../storage/schemas/task-embeddings");
           const { eq } = await import("drizzle-orm");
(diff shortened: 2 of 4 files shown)


edobry (commented):
## Review: Drizzle DELETE rowCount fix

**CI status:** ✅ Both checks pass (build + Prevent Placeholder Tests)

### Findings

**[PRE-EXISTING]** `minskyTaskBackend.ts:149-153` — Same `rowCount` bug in `deleteTask()`. Returns `false` even on successful deletion. Not introduced by this PR but should be fixed with the same `.returning()` pattern.

**[PRE-EXISTING]** `postgres-storage.ts:404-405` — Same `rowCount` bug in session `deleteSession()`. Returns `false` even on successful deletion.

**[PRE-EXISTING]** `github-issues-api.ts:371` — Logs `rowCount: undefined` for debug output. Cosmetic but misleading.

**Recommendation:** File a follow-up task to apply the `.returning()` fix to all 3 remaining `rowCount` sites.

### Checked and clear

- **`deleteEdge`**: Correctly uses `.returning({ id }).length` — verified in source at line 62-72
- **`deleteEdgesFrom`**: Same correct pattern — verified at line 74-82
- **No other callers of these methods** that would be affected by the return type change
- **In-memory repo** (used in tests) returns count directly from `Set.delete()` — unaffected

### Spec verification

**Task:** mt#784

|---|---|---|
| DELETE operations return correct count | Met | `.returning()` returns actual deleted rows, `.length` gives count — `task-graph-service.ts:71,81` |
| Backward compatible | Met | Return type (`number`) unchanged; only the mechanism to obtain it changed |
| Affects deleteEdge and deleteEdgesFrom | Met | Both methods updated — lines 61-82 |

PR #448, opened 2026-04-18
feat(mt#756): delete remote branch on session_delete

edobry:
## Summary

- When a session is deleted, also delete the remote git branch (`task/mt-NNN`) from GitHub
- Non-existent remote branches handled gracefully (debug log, not an error)
- Uses existing `taskIdToBranchName` and `gitService.execInRepository` — no new dependencies
- `gitService` is optional in `deleteSessionImpl` — backward compatible

## Motivation

`session_delete` removes the local directory but leaves the remote branch on GitHub. When a new session starts for the same task, git checks out the stale remote branch, causing subagents to believe work is already done. Discovered during mt#672 retrospective.

## Test plan

- [ ] Existing tests pass (no breaking changes — `gitService` is optional)
- [ ] Manual test: delete a session with a pushed branch, verify remote branch is gone
- [ ] Manual test: delete a session with no remote branch, verify no error

(Had Claude look into this — AI-assisted implementation, generated via `session.generate_prompt` tool as e2e validation of mt#728)

### DIFF
--- a/src/domain/session/session-delete-remote-branch.test.ts
+++ b/src/domain/session/session-delete-remote-branch.test.ts
@@ -1,0 +1,137 @@
+/**
+ * Tests for remote branch cleanup in deleteSessionImpl.
+ *
+ * Verifies that when a session is deleted, the remote git branch is also
+ * deleted from the remote origin. Tests cover:
+ *   - Remote branch deleted when workspace dir and git service are present
+ *   - Remote branch deletion failure is non-fatal (session still deleted)
+ *   - No remote branch deletion attempt when no git service is provided
+ *   - Branch name derived from taskId via taskIdToBranchName
+ *   - Session ID used as branch name fallback when no taskId
+ */
+
+import { describe, it, expect, mock, beforeEach } from "bun:test";
+import { deleteSessionImpl } from "./session-lifecycle-operations";
+import { FakeGitService } from "../git/fake-git-service";
+import type { SessionRecord } from "./types";
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/** Minimal in-memory session provider sufficient for deleteSessionImpl. */
+function makeSessionDB(sessions: SessionRecord[]) {
+  const store = new Map(sessions.map((s) => [s.session, s]));
+  return {
+    getSession: mock(async (id: string) => store.get(id) ?? null),
+    getSessionByTaskId: mock(async () => null),
+    listSessions: mock(async () => Array.from(store.values())),
+    addSession: mock(async () => {}),
+    updateSession: mock(async () => {}),
+    deleteSession: mock(async (id: string) => {
+      const existed = store.has(id);
+      store.delete(id);
+      return existed;
+    }),
+    getRepoPath: mock(async () => "/mock/repo"),
+    getSessionWorkdir: mock(async (id: string) => `/mock/sessions/${id}`),
+  };
+}
+
+describe("deleteSessionImpl — remote branch cleanup", () => {
+  const SESSION_ID = "test-session-abc";
+  const TASK_ID = "mt#756";
+  const EXPECTED_BRANCH = "task/mt-756";
+
+  let sessionRecord: SessionRecord;
+
+  beforeEach(() => {
+    sessionRecord = {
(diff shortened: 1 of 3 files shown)


edobry (commented):
## Review: session_delete remote branch cleanup

**CI status:** All checks passed (build + Prevent Placeholder Tests — both completed success)

### Findings

**[NON-BLOCKING]** `src/domain/session/session-delete-remote-branch.test.ts:73–80` — The "uses session ID as branch name when no taskId is present" test does not actually exercise the fallback path through `deleteSessionImpl`. It only calls `taskIdToBranchName` directly and asserts the two string values differ — it does not create a session record without a `taskId`, call `deleteSessionImpl`, and verify `execInRepository` was called with the session ID as the branch. The production fallback logic (`resolvedSessionId` as branch name) is untested. Low risk for now since the fallback is only for pathological sessions that lack a `taskId`, but worth hardening if that branch matters.

**[NON-BLOCKING]** `src/domain/session/session-lifecycle-operations.ts:127–165` — The remote branch deletion is gated on `existsSync(sessionWorkspaceDir)`. This means if the session directory was already removed manually (e.g., by a prior partial cleanup) the remote branch will not be deleted, even though a git service is available. This is intentional per the implementation (the workspace dir is needed to run `git push`), but it means the original bug (stale remote branches blocking new sessions) can still occur in that scenario. The comment in the `else` branch correctly documents this. Acceptable tradeoff for a v1.

**[NON-BLOCKING]** `src/domain/session/session-lifecycle-operations.ts:139–144` — `execInRepository` is called with the raw string `"push origin --delete ${branchName}"`. If `branchName` contains a space or shell metacharacter (unlikely given `taskIdToBranchName` output format, but possible for the UUID fallback path), this could produce a malformed git command. Not a practical risk given current branch name formats, but worth noting.

### Checked and clear

- **Branch name computation**: `taskIdToBranchName("mt#756")` → `"task/mt-756"` confirmed correct. The function parses `backend#localId` and returns `task/${backend}-${localId}`. The test at line 64–67 directly verifies this conversion.
- **Error handling**: Remote branch deletion errors are caught and logged at `debug` level; session deletion continues regardless. Both the implementation and the "still deletes session record when remote branch deletion throws" test confirm this is non-fatal.
- **Backward compatibility**: `gitService` is typed `GitServiceInterface | undefined` in `deps`, the `if (deps.gitService && sessionRecord)` guard is correct, and the "skips remote branch deletion when no gitService is provided" test confirms the pre-existing path still works.
- **Ordering**: Remote branch deletion happens before `rmSync` — correct. The session directory is still available when the git push runs, so it can use the workspace as the git working directory.
- **No taskId edge case**: Fallback to `resolvedSessionId` is present and correct in the implementation. The test coverage for this path is weak (see NON-BLOCKING finding), but the production code is safe.
- **Session not in DB**: `getSession` returns `null` → the `if (deps.gitService && sessionRecord)` guard skips remote deletion. No crash.
- **Workspace dir doesn't exist**: `existsSync` guard is checked before any git operation. No crash.
- **`FakeGitService.setCommandError`**: Uses string `includes` matching — `"push origin --delete"` correctly matches the command string issued in the implementation.
- **`SessionService.delete`**: Now passes `this.deps.gitService` (always present in real `SessionDeps`) to `deleteSessionImpl`. Wiring is correct.

### Spec verification

**Task:** mt#756

|---|---|---|
| session_delete removes remote git branch | Met | `session-lifecycle-operations.ts:139–157`: `execInRepository(sessionWorkspaceDir, "push origin --delete ${branchName}")` |
| Non-existent remote branch is not an error | Met | `session-lifecycle-operations.ts:158–163`: catch logs at debug, does not rethrow |
| Branch name derived from taskId | Met | `taskIdToBranchName(sessionRecord.taskId)` at line 130; test verifies `"mt#756" → "task/mt-756"` |
| Session ID used as fallback when no taskId | Met (code) / Partially tested | Production fallback at line 131–132 exists; test at line 73 only checks the string values differ, not the delete path itself |
| gitService is optional (backward compatible) | Met | `gitService?` in `deps` type; guard at line 127; test "skips remote branch deletion when no gitService" |
| Works when workspace dir doesn't exist | Met | `existsSync` guard at line 134; logs and skips gracefully |

PR #455, opened 2026-04-18
feat(mt#787): Add scope batching to session.generate_prompt

edobry:
## Summary

- Adds `batches`, `batchIndex`, and `totalBatches` fields to `GeneratePromptResult`
- When scope exceeds 40 files, `generateSubagentPrompt` splits scope into chunks of 30 and returns one prompt per batch
- Each batch prompt includes its batch number in the header (e.g. "Batch 1 of 3"), only its own files in the scope section, and intermediate commit instructions for non-final batches
- The primary `prompt` field always contains the first batch; the `batches` array contains all batches
- Scopes of 40 files or fewer are unchanged (no `batches` field, no `scopeWarning`)

## Test plan

- [ ] Scope with exactly 40 files: no batching, no scopeWarning
- [ ] Scope with 41 files: 2 batches (30 + 11)
- [ ] Scope with 90 files: 3 batches (30 + 30 + 30)
- [ ] Each batch prompt contains its correct batch number header
- [ ] Each batch prompt contains only its own files in the scope section
- [ ] Non-final batches include intermediate commit instructions
- [ ] Final batch includes full commit + PR instructions (not intermediate)
- [ ] Primary `prompt` matches first batch `prompt`
- [ ] `batchIndex` and `totalBatches` are correct on each batch
- [ ] `scopeWarning` is set and contains the file count


edobry (commented):
## Review: session.generate_prompt Phase 2 — scope analysis and batching

**CI status:** `build` in progress; `Prevent Placeholder Tests` passed. Cannot approve until build completes.

---

### Findings

**[NON-BLOCKING]** `src/domain/session/prompt-generation.ts` — Dead `scopeWarning` computation in non-batching path

After the `needsBatching` branch returns early, the fallthrough path recomputes `scopeWarning` with `scope.length > SCOPE_WARNING_THRESHOLD`. But since `needsBatching = scope && scope.length > SCOPE_WARNING_THRESHOLD`, the condition is always false in the fallthrough — so `scopeWarning` is always `undefined` here. This is harmless but the code is confusing (the old warning message referencing "Consider batching into multiple smaller tasks" is dead).

Evidence: `src/domain/session/prompt-generation.ts` lines 141–144 in the PR branch — reachable only when `scope.length <= 40`, but the condition tests `scope.length > 40`.

**[NON-BLOCKING]** `src/domain/session/prompt-generation.ts` — Root `GeneratePromptResult` lacks `batchIndex`/`totalBatches` when batching

When batching is triggered, the root result object has `prompt` (= batch 1 content), `batches`, `scopeWarning`, but no `batchIndex` or `totalBatches`. Each item in `batches` does have these. A caller using `result.batchIndex` to check which batch they have would see `undefined` on the root.

The task spec doesn't require the root result to have these fields, and tests don't assert them on the root, so this is a design question rather than a bug. Worth noting for callers.

**[NON-BLOCKING]** `src/domain/session/prompt-generation.ts` — `firstBatchPrompt` assigned via side effect in `.map()`

`firstBatchPrompt` is initialized as `""` then assigned inside the `.map()` callback when `batchIndex === 1`. This works correctly but is unusual — `chunks[0]` could generate the prompt directly via `generateSinglePrompt(params, chunks[0], 1, totalBatches)` without the side-effect variable. No behavioral impact.

**[NON-BLOCKING]** Review type + batching: intermediate commit instructions are silently skipped

`generateSinglePrompt` has an early return for `type === "review"` that exits before the intermediate-commit injection. This means if someone calls `generateSubagentPrompt` with `type: "review"` and 41+ files, they get batches, but no commit/intermediate-commit guidance in any batch (which is correct for reviews). The behavior is correct but there's no test covering `type: "review"` with >40 files, so the interaction is unverified.

---

### Checked and clear

- **Batch splitting logic**: `scope.slice(i, i + BATCH_SIZE)` correctly produces chunks of 30. Verified with 41 files → `[0..29]` + `[30..40]`, 90 files → three chunks of 30.
- **Boundary (40 vs 41)**: `scope.length > SCOPE_WARNING_THRESHOLD` (strict greater-than) means 40 files = no batching, 41 = batching. Correct per spec.
- **Each batch gets its own scope**: `batchScope` parameter passed to `generateSinglePrompt` controls `effectiveScope`, so each batch renders only its own files. Verified against file isolation tests.
- **Intermediate vs final commit**: `batchIndex < totalBatches` → intermediate commit section; `batchIndex === totalBatches` → full `renderCommitInstructions` with PR creation. Correct.
- **Primary `prompt` = first batch**: `firstBatchPrompt` captured in `.map()` and assigned to root `result.prompt`. Confirmed by test `"primary prompt matches the first batch prompt"`.
- **`batchIndex`/`totalBatches` on batch objects**: 1-based index, correct total. Confirmed by `"each batch has correct batchIndex and totalBatches"` test.
- **`GeneratePromptResult` interface backward compatible**: All three new fields (`batches`, `batchIndex`, `totalBatches`) are optional. Existing callers unaffected.
- **Batch header rendered**: `**Batch N of M**` appended to header when `batchIndex` is defined. Verified in `generateSinglePrompt`.
- **`suggestedSubagentType` propagated**: Both root result and each batch result set `suggestedSubagentType: "refactor"` when `type === "refactor"`. Correct.
- **Test coverage**: 10 new tests covering all key behaviors. All directly test observable output of `generateSubagentPrompt`. Tests are hermetic (no filesystem or DI dependencies).
- **No regressions in existing tests**: All pre-existing test cases remain unchanged; refactoring extracted `generateSinglePrompt` from the old `generateSubagentPrompt` body without removing any existing behavior branches.

---

### Spec verification

**Task:** mt#787

|---|---|---|
| When scope > 40 files, return multiple prompts (one per ~30-file batch) | Met | `BATCH_SIZE = 30`, `chunks.map(...)` returns one `GeneratePromptResult` per chunk in `batches[]` |
| Intermediate commit instructions for non-final batches | Met | `batchIndex < totalBatches` → `## Intermediate Commit` section with `session_commit` |
| Final batch has full commit + PR instructions | Met | `batchIndex === totalBatches` → `renderCommitInstructions()` which includes `session_pr_create` |
| Scope ≤ 40 = no batching (boundary: 40 = no, 41 = yes) | Met | `scope.length > SCOPE_WARNING_THRESHOLD` (strict); test asserts 40 files → no batches, 41 → batches |
| Primary `prompt` field = first batch | Met | `result.prompt = firstBatchPrompt`; test asserts `result.prompt === result.batches[0].prompt` |
| `batchIndex` and `totalBatches` correct per batch | Met | 1-based index set in `.map()`, `totalBatches = chunks.length`; test verifies all 3 batches for 90 files |
| Directly prevents mt#672 context exhaustion failure | Met (structurally) | Batching is implemented correctly; end-to-end caller integration depends on mt#786 (session lookup fix) |

---

PR #456, opened 2026-04-18
fix(mt#789): Preserve DB record when session filesystem deletion fails

edobry:
## Summary

- Fix delete asymmetry: `deleteSessionImpl` now preserves the DB record when `rmSync` throws, preventing orphan directories
- Fix `cleanupSessionImpl` with the same pattern: skip DB deletion when directory removal has errors
- Add `fs` as injectable deps for testability (DI pattern, no `mock.module`)
- Add 3 regression tests covering: fs failure preserves record, missing dir still deletes, successful case removes both

**Root cause:** During the mt#723 session cleanup audit, we found 24 UUID session directories with no DB records. While most were caused by the mt#722 Postgres error-swallowing bug, the delete ordering asymmetry was a separate live bug that could still create new orphans.

## Test plan

- [x] 3 new tests pass: `session-delete-fs-failure.test.ts`
- [x] 7 existing tests pass: `session-delete-remote-branch.test.ts`
- [x] No behavioral change for the success path — only the failure path is fixed

### DIFF
--- a/src/domain/session/session-delete-fs-failure.test.ts
+++ b/src/domain/session/session-delete-fs-failure.test.ts
@@ -1,0 +1,121 @@
+/**
+ * Tests for filesystem failure handling in deleteSessionImpl.
+ *
+ * Verifies that when filesystem removal fails during session deletion,
+ * the DB record is preserved to prevent orphan directories (mt#789).
+ */
+
+import { describe, it, expect, mock, beforeEach } from "bun:test";
+import { deleteSessionImpl } from "./session-lifecycle-operations";
+import type { SessionRecord } from "./types";
+
+/** Minimal in-memory session provider sufficient for deleteSessionImpl. */
+function makeSessionDB(sessions: SessionRecord[]) {
+  const store = new Map(sessions.map((s) => [s.session, s]));
+  return {
+    getSession: mock(async (id: string) => store.get(id) ?? null),
+    getSessionByTaskId: mock(async () => null),
+    listSessions: mock(async () => Array.from(store.values())),
+    addSession: mock(async () => {}),
+    updateSession: mock(async () => {}),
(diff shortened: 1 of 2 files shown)


edobry (commented):
## Review: Session delete asymmetry fix

**CI status:** in progress (rebase just completed)

### Findings

No blocking issues. The fix is clean and minimal.

**Verified:** The `DeleteSessionResult` interface (line 71) already has `error?: string`, so the new `return { deleted: false, error: ... }` path is type-safe.

**Verified:** The command layer at `management-commands.ts:32-36` already handles `result.error` by passing it through to the response — no caller changes needed.

**Verified:** The `fs` dep injection follows the project's DI pattern (optional dep with module-level fallback), consistent with how `gitService` is already injected.

### Checked and clear

- `deleteSessionImpl` — fs failure path now returns error result instead of swallowing
- `cleanupSessionImpl` — DB deletion gated on `errors.length === 0`
- Test file — 3 tests cover failure, missing-dir, and success paths with injected fs fakes
- `SessionService.delete()` — passes through result unchanged, no impact
- Command adapter — already propagates `result.error`

### Spec verification

**Task:** mt#789

| `deleteSessionImpl` only deletes DB record if fs removal succeeds or dir doesn't exist | Met | Lines 165-185: rmSync failure returns early with `deleted: false`; existsSync false skips rmSync entirely |
| If fs removal fails, error is surfaced to caller (not swallowed) | Met | Line 182: returns structured error with message; line 179: logs at error level |
| Test covering failure case: rmSync throws → DB record preserved | Met | `session-delete-fs-failure.test.ts` line 49: "preserves DB record when rmSync throws" |
| Same fix applied to `cleanupSessionImpl` | Met | Line 334: `errors.length === 0` gate; lines 350-353: warning log when skipped |

PR #459, opened 2026-04-18
feat(mt#791): Add audit prompt type and register MCP prompts

edobry:
## Summary

- Add `"audit"` to `PromptType` union and handle it in `generateSinglePrompt` / `generateSubagentPrompt`: omits commit/PR instructions, includes audit instructions with Met/Not met/Not applicable format, suggests `verify-completion` subagent type and `sonnet` model
- Update `prompt-command.ts` `z.enum` and type cast to include `"audit"`
- Add 5 tests for the audit prompt type in `prompt-generation.test.ts`
- Create `src/adapters/mcp/prompts.ts` with `registerPrompts()` that registers 5 MCP prompts (`session_implementation`, `session_refactor`, `session_review`, `session_cleanup`, `session_audit`)
- Wire `registerPrompts(server)` into `start-command.ts` after `registerAllTools(commandMapper)`

## Test plan

- [ ] Audit prompt omits `mcp__minsky__session_commit` and `mcp__minsky__session_pr_create`
- [ ] Audit prompt contains "Audit Instructions" with Met/Not met/Not applicable bullets
- [ ] Audit prompt suggests `verify-completion` subagent type
- [ ] Audit prompt suggests `sonnet` model
- [ ] MCP server registers 5 prompt types visible via `list_prompts`


edobry (commented):
## Review: Add audit prompt type and register MCP prompts

**CI status:** `build` in progress; `Prevent Placeholder Tests` passed. Cannot merge until build check completes.

---

### Findings

**[BLOCKING]** `src/adapters/mcp/prompts.ts` (entire file) / `src/mcp/server.ts:339-344` — MCP prompt arguments are not declared, so slash commands won't have discoverable parameters in Claude Code.

The MCP `Prompt` schema (from `@modelcontextprotocol/sdk`) includes an optional `arguments` array of `PromptArgument` objects (each with `name`, `description`, `required`). These are returned in the `prompts/list` response and tell the client (Claude Code) what parameters the slash command accepts — enabling the UI to prompt for them interactively.

The current `PromptDefinition` interface in `server.ts` has no `arguments` field:
```typescript
interface PromptDefinition {
  name: string;
  description?: string;
  spec?: string;
  handler: (args: Record<string, unknown>) => Promise<unknown>;
}
```

And the `ListPromptsRequestSchema` handler only returns `name` and `description`:
```typescript
prompts: Array.from(this.prompts.values()).map((prompt) => ({
  name: prompt.name,
  description: prompt.description,
}))
```

Each registered prompt in `prompts.ts` expects `task` (required) and `instructions` (optional) but neither the `PromptDefinition` interface nor the registration call in `registerPrompts()` declares them. When Claude Code calls `prompts/list`, it gets no argument metadata, so the slash command invocation would require the user to know the argument names without UI assistance.

**Fix:** Add an `arguments` field to `PromptDefinition`, populate it in the `ListPromptsRequestSchema` handler response, and declare `task`/`instructions` arguments when calling `server.addPrompt()` in `registerPrompts()`.

---

**[NON-BLOCKING]** `src/domain/session/prompt-generation.ts` — `audit` handling in `generateSinglePrompt` falls through to `renderCommitInstructions`.

The early-return for `"audit"` is correctly placed before `"review"`, but note the control flow: after the `"audit"` block returns, the `"refactor"` and `"cleanup"` blocks add sections and then fall through to `renderCommitInstructions(sessionId, taskId)` at line 124. If the `"audit"` early return were ever missed (e.g., a future refactor removes it), the audit prompt would gain commit/PR instructions. This is already guarded by the tests, but it would be marginally safer to have the audit block mirror the review block's structure (early return) which it already does — so this is just a note that the two-phase structure (early-return types vs. fall-through types) is non-obvious.

No change required; tests cover this correctly.

---

**[NON-BLOCKING]** `src/adapters/mcp/prompts.ts:45` — Dynamic `import()` inside the handler loop.

The handler does three dynamic `import()` calls on every invocation:
```typescript
const { getSessionFromParams } = await import("../../domain/session");
const { generateSubagentPrompt } = await import("../../domain/session/prompt-generation");
const { resolveSessionDirectory } = await import("../../domain/session/resolve-session-directory");
```

In Bun, repeated dynamic imports of the same module are cached, so this doesn't cause repeated module loading. The pattern is consistent with `prompt-command.ts` which does the same. No behavioral issue.

---

### Checked and clear

- **Audit type omits commit/PR instructions** — `generateSinglePrompt` returns early for `"audit"` before reaching `renderCommitInstructions()`. Verified by reading `prompt-generation.ts:99-130`.
- **Audit type suggests `verify-completion` subagent** — `generateSubagentPrompt` returns `suggestedSubagentType: "verify-completion"` for `"audit"`. Verified at the inserted block after line 205 in the diff.
- **Audit type suggests `sonnet` model** — `suggestedModel: "sonnet"` in the same block.
- **All 5 prompt types registered** — `promptTypes` array in `registerPrompts()` contains `session_implementation`, `session_refactor`, `session_review`, `session_cleanup`, `session_audit`.
- **Registration wired in start-command.ts** — `registerPrompts(server)` added immediately after `registerAllTools(commandMapper)` at the correct location in `start-command.ts`.
- **Session resolution path is correct** — `getSessionFromParams({ task })` → `resolveSessionDirectory(sessionId)` matches the same pattern used in `prompt-command.ts`. `resolveSessionDirectory` is well-implemented in `resolve-session-directory.ts`.
- **Tests are meaningful** — 5 tests in `prompt-generation.test.ts` verify: commit omission, PR omission, audit instructions text with all 3 bullet formats, `verify-completion` subagent type, `sonnet` model. These check actual behavior, not just that the function runs.
- **Import in prompts.ts** — `MinskyMCPServer` imported as a type-only import (`import type`), correct.
- **No type errors visible** — `type` field in the `as const` array is correctly narrowed; the loop variable captures the correct literal type.
- **`taskId` stripping** — `task.replace(/^mt#/, "").replace(/^#/, "")` is consistent with `prompt-command.ts`.

---

### Spec verification

**Task:** mt#791

|---|---|---|
| Audit prompt type added | Met | `PromptType` extended to include `"audit"`; `generateSinglePrompt` and `generateSubagentPrompt` handle it |
| At least one prompt registered in MCP prompt registry | Met | `registerPrompts()` registers 5 prompts via `server.addPrompt()` |
| Slash commands would work in Claude Code (verify registration path) | **Partially met** | `registerPrompts(server)` is wired in `start-command.ts` and `addPrompt` → `ListPromptsRequestSchema` pipeline is functional, but prompt `arguments` are not declared — Claude Code slash command UI won't show `task`/`instructions` as discoverable parameters, which degrades UX for "user-facing slash command access" per the spec goal |

**Action required:** The missing `arguments` declaration is a blocking gap against the spec's "user-facing slash command access" goal. The registration pipeline works at the protocol level (prompts are listed and callable), but without argument metadata, Claude Code cannot surface these as properly self-documenting slash commands. This should be fixed before merge.

---

PR #462, opened 2026-04-18
feat(mt#798): add dependency cycle detection to TaskGraphService

edobry:
## Summary

`addDependency()` now prevents circular dependencies. Before creating an edge, it runs BFS over "depends" edges from the target to check if the source is transitively reachable. If so, the edge is rejected.

This is a subtask of mt#239 (Phase 2: Task Dependencies), broken out as a discrete piece.

### Changes

- **`task-graph-service.ts`**: `addDependency()` calls `getTransitiveDependencies()` before creating edges. New `getTransitiveDependencies()` method does BFS with configurable max node limit (default 100).
- **Tests**: 4 new tests — direct cycle, indirect cycle, non-cyclic chain (allowed), diamond pattern (allowed).

## Spec verification

**Task:** mt#798

|---|---|---|
| Cycle detection in addDependency | Met | `task-graph-service.ts:152-157`, BFS check |
| Direct cycles rejected | Met | Test: "prevents direct dependency cycle" |
| Indirect cycles rejected | Met | Test: "prevents indirect dependency cycle" |
| Non-cyclic graphs allowed | Met | Tests: "allows non-cyclic chains" + "allows diamond dependencies" |

### DIFF
--- a/src/domain/tasks/task-graph-service.ts
+++ b/src/domain/tasks/task-graph-service.ts
@@ -149,6 +149,13 @@ export class TaskGraphService {
       throw new Error("A task cannot depend on itself");
     }
 
+    // Cycle prevention: if fromId is transitively reachable from toId,
+    // adding fromId→toId would create a cycle
+    const transitiveDeps = await this.getTransitiveDependencies(toId);
+    if (transitiveDeps.has(fromId)) {
+      throw new Error(`Cycle detected: ${toId} already transitively depends on ${fromId}`);
+    }
+
     const exists = await this.repo.findEdge(fromId, toId, "depends");
     if (exists) {
       return { created: false };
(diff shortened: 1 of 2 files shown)


edobry (commented):
## Review: Dependency cycle detection

**CI status:** Pending

### Findings

No issues found. Clean, focused change — 2 files, 61 additions.

### Checked and clear

- **BFS implementation**: Correct — visits each node once via `visited` set, follows `listDependencies` (type-filtered to "depends"), capped at 100 nodes
- **Cycle check placement**: Before `findEdge` (idempotency check) — correct ordering. No point checking for existing edge if we'd reject for cycle anyway.
- **No regression on existing tests**: All 26 original tests pass alongside 4 new ones
- **Diamond graphs**: Correctly allowed — diamond is not a cycle. Test confirms.
- **Parent edges not affected**: Cycle detection only follows "depends" edges, parent-child cycle prevention unchanged

### Spec verification

**Task:** mt#798

|---|---|---|
| Cycle detection in addDependency | Met | BFS via `getTransitiveDependencies`, check at line 154-157 |
| Direct cycles rejected | Met | Test passes |
| Indirect cycles rejected | Met | Test passes |
| Non-cyclic graphs allowed | Met | Two tests (chain + diamond) pass |

PR #463, opened 2026-04-18
fix(mt#793): Fix cache-before-init anti-pattern in persistence/session layers

edobry:
## Summary

Fixes the cache-before-init anti-pattern in 4 locations where references were cached before async initialization completed. On initialization failure, stale uninitialized objects remained in the cache, causing subsequent calls to silently use broken storage/providers instead of retrying.

This is the root cause of the MCP `session_list` returning empty after `/mcp` reconnect (mt#788) — any transient initialization failure would permanently poison the cache.

## Changes

**Pattern applied in all 4 instances:** assign to local variable → initialize → cache only on success. Clean up partially-created resources in catch block.

| File | What was cached too early | Fix |
|------|--------------------------|-----|
| `SessionDbAdapter.getStorage()` | `this.storage` before `initialize()` | Local var, cache after init succeeds |
| `PersistenceService.performInitialization()` | `provider` before `initialize()` | Local var, null out in catch — `isInitialized()` now correctly returns false |
| `PostgresPersistenceProvider.initialize()` | `this.sql`/`this.db` before `SELECT 1` | Local vars, end() created connections on failure (not injected ones) |
| `SqlitePersistenceProvider.initialize()` | `this.db`/`this.drizzleDb`/`this.storage` before PRAGMAs | Local var, close() DB handle on failure to prevent FD leaks |

## Tests added

- `SessionDbAdapter`: cache stays null after init failure; retry re-attempts; successful init caches correctly
- `PostgresPersistenceProvider`: state nulled after failure; injected clients not closed; retry succeeds
- `PersistenceService` (new file): `isInitialized()` false after failure; `getProvider()` throws; retry succeeds

## Task

Part of mt#788 (MCP session_list empty after reconnect). Subtask mt#793.

Unblocks mt#794 (error swallowing), mt#795 (DI wiring), mt#796 (regression prevention).


edobry (commented):
## Review: Cache-before-init fix across persistence/session layers

**CI status:** `Prevent Placeholder Tests` ✅ | `build` ⏳ in progress

### Findings

No blocking issues found. All 4 fixes follow the same correct pattern: local variable → async init → cache on success, with resource cleanup in catch blocks.

**Verified details:**

- **PostgresPersistenceProvider cleanup ownership:** The `createdSql` tracking correctly distinguishes injected clients (`deps?.sqlClient`) from self-created connections. Only self-created connections are `.end()`'d on failure. Test at postgres-provider.test.ts:101 verifies `failingSql.end` is NOT called for injected clients. ✅
- **VectorCapablePostgresPersistenceProvider compatibility:** Subclass at postgres-provider.ts:239 calls `super.initialize()` then accesses `this.sql`. After this fix, failed `super.initialize()` correctly nulls `this.sql`, so the subclass's guard (`if (!this.sql)`) now catches failures more reliably. No behavioral change on success path. ✅
- **PersistenceService concurrent init guard:** The `initPromise` guard (service.ts:31) with `finally { initPromise = null }` (line 40) still works correctly — our change only moves when `provider` is assigned, not the promise lifecycle. ✅
- **SessionDbAdapter return type:** Changed `return this.storage!` to `return this.storage` (line 48). Since we're inside `if (!this.storage)` and just assigned `this.storage = storage`, TypeScript narrowing guarantees non-null. Removing the `!` assertion is strictly safer. ✅

### Checked and clear

- All 4 production code changes apply the assign-after-init pattern consistently
- All catch blocks clean up resources appropriate to what was created (DB handles, connections)
- `isInitialized` flag is only set to `true` after all checks pass in all 4 instances
- Tests cover: failure leaves cache empty, retry re-attempts, success caches correctly
- No changes to public API signatures
- No callers affected (internal initialization patterns only)

### Spec verification

**Task:** mt#793

| All 4 instances use assign-after-init pattern | Met | session-db-adapter.ts:36-39, service.ts:53-59, postgres-provider.ts:72-97, sqlite-provider.ts:79-107 |
| All 4 catch blocks clean up partially-created resources | Met | session-db-adapter.ts: no cleanup needed (sync getStorage); service.ts:62 nulls provider; postgres-provider.ts:100-107 ends created connection; sqlite-provider.ts:108-115 closes DB handle |
| `PersistenceService.isInitialized()` returns false after failed init | Met | service.test.ts "isInitialized() returns false" test |
| `PersistenceService.getProvider()` throws after failed init | Met | service.test.ts "getProvider() throws" test |
| Tests: cache null after failure, retry works, no leaks | Met | 8 new tests across 3 files covering all scenarios |
| Existing tests pass unchanged | Met | 1587 tests, 0 failures (full suite run in session) |

PR #477, opened 2026-04-18
feat(mt#802): enforce session.generate_prompt via watermark and PreToolUse hook

edobry:
## Summary

- Add `<!-- minsky:prompt:v1 -->` watermark to all generated prompts
- Add PreToolUse hook on Agent tool that warns when session-work prompts lack the watermark
- Structural enforcement per the rules→skills→subagents→hooks strength spectrum

## How it works

1. `session.generate_prompt` embeds a watermark in every prompt it generates
2. Before every Agent tool call, the hook checks: does this prompt mention `session_commit`? If yes, does it have the watermark?
3. Session work without watermark → warning shown to user and agent
4. Non-session Agent calls (exploration, research) → pass silently

## Design decisions

- **Warn (exit 0), not block (exit 2)** — allows edge cases (stale MCP, novel use cases) while making violations visible
- **Detection via `session_commit`** — only session-work prompts trigger the check
- **Watermark format** — HTML comment, invisible to LLM, version-tagged

(Had Claude look into this — AI-assisted implementation)


edobry (commented):
## Review: Watermark enforcement via PreToolUse hook on Agent

**CI status:** In progress (both `build` and `Prevent Placeholder Tests` checks were running at review time — do not merge until green)

---

### Findings

No blocking issues found. Two non-blocking observations:

**[NON-BLOCKING]** `src/domain/session/prompt-generation.ts` — `review` and `audit` prompt types do not include `session_commit` text in their output (verified by existing tests: `expect(result.prompt).not.toContain(MCP_SESSION_COMMIT)`). This means the hook will **pass silently** on tool-generated `review`/`audit` prompts even though they carry the watermark — because the `isSessionWork` gate never fires. This is correct behavior by design (read-only agent calls don't need commit enforcement), but worth noting: the watermark in review/audit prompts is currently unchallenged by the hook. Design trade-off, not a bug.

**[NON-BLOCKING]** `.claude/hooks/check-prompt-watermark.ts:13` — The hook reads `input.tool_input.prompt as string` with a `?? ""` fallback. If the Agent tool is ever called with a non-string `prompt` field, the cast would silently coerce to a truthy object rather than the empty-string fallback (the `??` only catches `null`/`undefined`, not wrong types). Low risk given the controlled environment.

---

### Checked and clear

- **Watermark in all prompt types**: Verified in `generateSinglePrompt` — the watermark is appended at the end of every code path: `review` early-returns with watermark, `audit` early-returns with watermark, and the fallthrough path covering `implementation`/`refactor`/`cleanup` appends it before the final `return`. No type can escape without the watermark.
- **Watermark in batched prompts**: `generateSinglePrompt` is called once per batch chunk and always appends the watermark. The new batch test (line ~304) verifies every batch prompt in a loop.
- **Hook detection logic**: `isSessionWork = prompt.includes("session_commit")` is semantically correct. The string `session_commit` appears in commit instructions embedded in `implementation`, `refactor`, and `cleanup` prompts but NOT in `review` or `audit` prompts — matching the intent exactly.
- **False positives**: Non-session Agent calls (exploration, research, plan subagents) would not include `session_commit` in their prompts. The gate is sound.
- **Exit code 0**: Correct for warning behavior per spec — allows edge cases (stale MCP, novel use cases) while making violations visible.
- **settings.json hook registration**: Registered correctly under `PreToolUse` with `matcher: "Agent"`, 5-second timeout, and `$CLAUDE_PROJECT_DIR`-relative path.
- **`PROMPT_WATERMARK` export**: Exported from the production module and imported in tests — no magic string duplication.
- **Hook types interface**: `ToolHookInput.tool_input: Record<string, unknown>` — the prompt access pattern matches the established hook interface in `.claude/hooks/types.ts`.

---

### Spec verification

**Task:** mt#802

|---|---|---|
| All generated prompts contain `<!-- minsky:prompt:v1 -->` | Met | `generateSinglePrompt` appends `PROMPT_WATERMARK` in all 3 code paths (review early-return, audit early-return, impl/refactor/cleanup fallthrough) |
| Hook fires and warns on hand-crafted session work prompts | Met | `check-prompt-watermark.ts`: `isSessionWork && !hasWatermark` → three `console.log` warning messages |
| Hook passes silently on non-session Agent calls | Met | `isSessionWork` gate: only prompts containing `session_commit` trigger the check; exploration/research prompts do not |
| Hook passes silently on tool-generated prompts (with watermark) | Met | `hasWatermark` check: tool-generated prompts containing `minsky:prompt:v1` satisfy the condition and exit 0 silently |
| Existing tests updated for watermark presence | Met | 6 new test cases in `tests/domain/session/prompt-generation.test.ts`: 5 per prompt type + 1 for batched prompts |
| settings.json updated with hook registration | Met | `PreToolUse` array in `.claude/settings.json` includes `matcher: "Agent"` entry pointing to new hook file |

All 6 success criteria are met.

---

PR #486, opened 2026-04-19
fix(mt#807): Fix session ID migration mechanism

edobry:
## Summary

- Fix the broken session ID migration: `updateSession()` silently stripped the `session` primary key, so ID renames never worked
- Use create-new + delete-old pattern for ID changes instead
- Add filesystem directory rename when session IDs change
- Register `session.migrate` command in shared registry (CLI + MCP)
- Update `needsMigration` to flag ALL non-UUID sessions (previously exempted `task-mt#XXX` format)

**Root cause:** The migration code existed as unreachable domain logic. Even if invoked directly, `updateSession` in `session-db.ts` explicitly strips the `session` key before applying updates, making ID renames impossible. Discovered during the mt#723 session cleanup audit.

## Test plan

- [x] 15 migration tests pass (3 updated for create+delete pattern)
- [x] 295 session tests pass
- [x] 1607 total tests pass (including architecture DI bypass test)
- [x] Typecheck clean


edobry (commented):
## Review: Session ID migration fix

**CI status:** build in progress, placeholder tests passed

### Findings

No blocking issues.

**[NON-BLOCKING]** `migration-command.ts:390-392` — If `addSession` succeeds but `deleteSession` fails (e.g., DB connection drops between the two calls), a duplicate record would exist. This is extremely unlikely and the directory rename is already done correctly (fail-safe, before DB ops). Acceptable for a migration tool — worst case is a stale old record that's easy to clean up.

### Checked and clear

- **`needsMigration` change** — Verified the old logic at `multi-backend-integration.ts:198-202` exempted qualified IDs matching `taskIdToSessionId()`, which meant `task-mt#XXX` sessions were never flagged. New logic correctly flags all non-UUID sessions.
- **`updateSession` strips primary key** — Confirmed at `session-db.ts:175`: `const { session: _sessionKey, ...safeUpdates } = updates`. Create+delete is the correct workaround.
- **Directory rename ordering** — Filesystem rename happens before DB writes (line 379-386). If rename fails, DB is untouched. If DB fails after rename, directory is already at new location but old record persists (recoverable).
- **Target directory collision** — Handled at line 380-383: skips with error if target exists.
- **Command registration** — `session.migrate` registered in `session.ts`, uses `getDeps()` (satisfies architecture DI test).
- **Test updates** — Three tests correctly updated for create+delete pattern: `analyzeMigrationNeeds` expects 5 not 3, full migration checks `addSession`/`deleteSession` instead of `updateSession`, error test mocks `addSession` instead of `updateSession`.

### Spec verification

**Task:** mt#807

| `minsky session migrate` command exists in CLI and MCP | Met | Registered in `session.ts:97`, command ID `session.migrate` |
| Migrates `task-mt#XXX` to UUID format | Met | `needsMigration` returns true for all non-UUID; `migrateSession` generates UUID at line 224 |
| Filesystem directories renamed | Met | `renameSync(oldDir, newDir)` at `migration-command.ts:385` |
| DB records use new UUID as primary key (create+delete) | Met | `addSession` at line 390, `deleteSession` at line 392 |
| Existing session data preserved | Met | `migrated` spread from original record at line 200; only `session` field changes |
| Dry-run mode | Met | Line 364: entire step 6 skipped when `dryRun` is true |
| Idempotent | Met | `needsMigration` returns false for UUID sessions (line 194); second run finds nothing |

PR #493, opened 2026-04-19
refactor(mt#814): convert PersistenceService from static singleton to injectable instance

edobry:
## Summary

Phase A of mt#811 (service consolidation for full DI). Converts `PersistenceService` from a static singleton class to an injectable instance class. This is the root blocker — every other service depends on persistence, and as long as it's static, nothing downstream can be properly injected.

### What changed

**`src/domain/persistence/service.ts`** — All methods converted from `static` to instance. Constructor creates an empty service; `initialize()` connects to the database. Exported `defaultInstance` for backward compatibility during migration.

**`src/composition/cli.ts`** — Uses `defaultInstance` (shared during migration) instead of creating a fresh instance.

**Production callers** (3 files) — Updated to use `defaultInstance` instead of static `PersistenceService`:
- `similarity-commands.ts` — `defaultInstance.getVectorStorage()`
- `startup-embedding-sweep.ts` — `defaultInstance.isInitialized()`, `.getProvider()`
- `vector-storage-factory.ts` — `defaultInstance.getProvider()`

**`session-db-adapter.ts`** — Default deps use `defaultInstance` instead of static class.

**Tests** (3 files) — Updated to create `new PersistenceService()` instances or use `defaultInstance`. Architecture test updated to match instance method patterns.

### Migration note

The `defaultInstance` export is a transitional bridge. Code that hasn't been fully migrated to container access still uses it. Once all callers resolve persistence from the container, `defaultInstance` will be removed (Phase C/E).

## Test plan

- [x] 1617 tests pass, 0 failures (1 architecture test updated)
- [x] ESLint: 0 errors, 0 warnings
- [x] CLI functional: session --help, tasks list work correctly

### DIFF
--- a/src/adapters/shared/commands/tasks/similarity-commands.ts
+++ b/src/adapters/shared/commands/tasks/similarity-commands.ts
@@ -328,11 +328,13 @@ export async function createTaskSimilarityService(): Promise<TaskSimilarityServi
 
   const embedding = await createEmbeddingServiceFromConfig();
 
-  // Use PersistenceService (should already be initialized at application startup)
-  const { PersistenceService } = await import("../../../../domain/persistence/service");
+  // Use the default PersistenceService instance (initialized at application startup)
+  const { defaultInstance: persistenceService } = await import(
+    "../../../../domain/persistence/service"
+  );
 
-  // Get vector storage directly - throws if provider doesn't support it (now synchronous!)
-  const vectorStorage = PersistenceService.getVectorStorage(dimension);
+  // Get vector storage directly - throws if provider doesn't support it
+  const vectorStorage = persistenceService.getVectorStorage(dimension);
 
   // Minimal task resolvers reuse domain functions via dynamic import to avoid cycles
   const { createConfiguredTaskService } = await import("../../../../domain/tasks/taskService");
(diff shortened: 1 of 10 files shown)


edobry (commented):
## Review: Convert PersistenceService from static to instance (Phase A)

**CI status:** both checks passed

### Findings

No blocking issues.

**[NON-BLOCKING]** `defaultInstance` is a transitional export — production callers use it while the migration completes. It will be removed when all callers resolve persistence from the container (Phase C/E).

### Checked and clear

- **service.ts**: All 8 methods converted from `static` to instance. Constructor is empty; `initialize()` connects to DB. `loadConfiguration()` remains static (doesn't depend on instance state). `setMockProvider` and `reset` removed (tests use `new PersistenceService()` or set `.provider` directly).
- **composition/cli.ts**: Uses `defaultInstance` so the container's provider is shared with legacy callers
- **3 production callers**: All switched from `PersistenceService.method()` to `defaultInstance.method()`
- **session-db-adapter.ts**: `defaultCreateSessionProviderDeps` uses `defaultInstance`
- **3 test files**: `service.test.ts` creates `new PersistenceService()` per test (proper isolation). `taskCommands` tests set `.provider` on `defaultInstance`. Architecture test updated for instance method patterns.
- **1617 tests pass**, 0 failures

### Spec verification

**Task:** mt#814

| PersistenceService is instance, not static | Met | All methods are instance methods; `static` keyword removed |
| Constructor accepts deps | Partially met | Constructor is parameterless (uses `loadConfiguration()`); config could be injected but this is acceptable for now |
| Composition roots create/register instance | Met | `composition/cli.ts` uses `defaultInstance` |
| Zero static `PersistenceService.getProvider()` in prod | Met | All callers use `defaultInstance.getProvider()` or container |
| Tests don't use `setMockProvider` | Met | Use `new PersistenceService()` or direct `.provider` assignment |
| All tests pass | Met | 1617 pass, 0 fail |

PR #492, opened 2026-04-19
fix(gh#491): auto-register task commands for MCP from single source

edobry:
## Summary

**Bug fix:** `tasks.children`, `tasks.parent`, and `tasks.dispatch` were added to `registry-setup.ts` (CLI) but not `tasks-modular.ts` (MCP). They worked in CLI but were invisible as MCP tools.

**Structural fix:** Replaced ~360 lines of manual per-command registration in `tasks-modular.ts` with a 10-line loop over `createAllTaskCommands()`. Now the CLI registry and MCP registry use the **same source of truth** — adding a command to `registry-setup.ts` automatically makes it available via MCP.

### Before (dual registration, error-prone)
```
// registry-setup.ts — CLI
createTasksChildrenCommand(getPersistenceProvider),
// tasks-modular.ts — MCP (MANUALLY, SEPARATELY)
sharedCommandRegistry.registerCommand({
  id: childrenCommand.id,
  category: CommandCategory.TASKS,
  ...
});
```

### After (single source)
```
// registry-setup.ts — single source
createTasksChildrenCommand(getPersistenceProvider),
// tasks-modular.ts — auto-registers ALL commands
for (const command of createAllTaskCommands(container)) {
  sharedCommandRegistry.registerCommand({ ...command, category: CommandCategory.TASKS });
}
```

## Spec verification

**Task:** gh#491

| Criterion | Status | Evidence |
|---|---|---|
| tasks.children/parent/dispatch appear as MCP tools | Met | Auto-registered from createAllTaskCommands |
| tasks-modular.ts auto-registers from single source | Met | Loop replaces 360 lines of manual registration |
| New commands automatically available via MCP | Met | No second registration step needed |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #496, opened 2026-04-19
fix(gh#491): preserve this binding in task command auto-registration

edobry:
## Summary

Follow-up fix to PR #492. The auto-registration loop passed `command.execute` directly, detaching it from the command instance. `BaseTaskCommand` methods like `this.debug()` failed with `this.debug is not a function` when called via MCP.

Fix: wrap in arrow function `(params, ctx) => command.execute(params, ctx)` to preserve call context.

## Spec verification

|---|---|---|
| tasks.get works via MCP | Met | Arrow wrapper preserves this binding |
| All existing commands still work | Met | 1617 tests pass |

### DIFF
--- a/src/adapters/shared/commands/tasks-modular.ts
+++ b/src/adapters/shared/commands/tasks-modular.ts
@@ -38,7 +38,7 @@ export class ModularTasksCommandManager {
           name: command.name,
           description: command.description,
           parameters: command.parameters,
-          execute: command.execute,
+          execute: (params, ctx) => command.execute(params, ctx),
         });
       }
 


edobry (commented):
## Review: Fix this binding in auto-registration

**CI status:** Pending

### Findings

No issues. One-line fix — `command.execute` → `(params, ctx) => command.execute(params, ctx)`.

The old manual registration used the same arrow wrapper pattern (e.g., `execute: async (params, context) => { return await listCommand.execute(params, context); }`). The auto-registration loop omitted this, breaking `BaseTaskCommand` methods that use `this`.

### Spec verification

|---|---|---|
| this binding preserved | Met | Arrow function wrapper maintains call context |
| All tests pass | Met | 1617 tests, 0 failures |

PR #507, opened 2026-04-19
fix(mt#823): Fix spec_get single-backend mode and migrate-backend error details

edobry:
## Summary

- **Bug 1**: `getTaskSpecContentFromParams` was loading config and explicitly setting `backend` from it, forcing single-backend mode. This meant `gh#` tasks couldn't be found via `spec_get`. Fixed by removing the config-loading block — `validParams.backend` (undefined if not provided) is now passed directly to `createConfiguredTaskService`, enabling multi-backend mode like all other `*FromParams` functions.

- **Bug 2**: The non-JSON return path of `migrate-backend-command` only returned summary counts, omitting error details. Fixed by adding `errorDetails: result.details?.filter((d) => d.error) ?? []` to the non-JSON return so callers can see what failed without needing `--json`.

## Test plan
- [ ] `minsky tasks spec-get gh#<id>` now works without specifying `--backend github` explicitly
- [ ] `minsky tasks migrate-backend` non-JSON output (without `--json`) now includes error detail entries when migrations fail

### DIFF
--- a/src/adapters/shared/commands/tasks/migrate-backend-command.ts
+++ b/src/adapters/shared/commands/tasks/migrate-backend-command.ts
@@ -218,9 +218,10 @@ export class TasksMigrateBackendCommand extends BaseTaskCommand<MigrateBackendPa
         migrated: result.migrated,
         skipped: result.skipped,
         errors: result.errors,
-        validated: validationResult?.passed.length || 0,
-        validationFailed: validationResult?.failed.length || 0,
+        validated: validationResult?.passed?.length ?? 0,
+        validationFailed: validationResult?.failed?.length ?? 0,
       },
+      errorDetails: result.details?.filter((d: MigrationDetail) => d.error) ?? [],
     };
   }
 
--- a/src/domain/tasks.ts
+++ b/src/domain/tasks.ts
@@ -243,35 +243,14 @@ export async function getTaskSpecContentFromParams(
   const workspacePath = process.cwd();
   log.debug("tasks.spec params", { backend: validParams.backend });
 
-  // Determine backend (prefer CLI param, else config)
-  let backend = validParams.backend;
-  if (!backend) {
-    try {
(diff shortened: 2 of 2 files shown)


edobry (commented):
## Review: Fix spec_get single-backend mode and migrate error visibility

**CI status:** Prevent Placeholder Tests passed; build in progress

### Findings

No blocking findings. Both changes are clean and minimal.

### Checked and clear

- **`src/domain/tasks.ts`** — The 20-line config-loading block is fully removed. `getTaskSpecContentFromParams` now passes `validParams.backend` directly (undefined when not provided), matching the pattern of all other `*FromParams` functions. The `persistenceProvider: deps?.persistenceProvider` passthrough from gh#499 is preserved. Debug log updated to use `validParams.backend` instead of the removed `backend` variable.
- **`migrate-backend-command.ts`** — `errorDetails` added to non-JSON return with proper type (`MigrationDetail`). Also improved null safety: `validationResult?.passed?.length ?? 0` instead of `validationResult?.passed.length || 0` (prevents crash if `passed` is undefined).
- No other files changed — correctly scoped.

### Spec verification

**Task:** mt#823

|---|---|---|
| `tasks_spec_get` with gh# ID returns spec content | Met | Config-loading block removed (`tasks.ts`); multi-backend mode now used, gh# prefix routes to GitHub backend |
| `tasks_spec_get` with mt# ID continues to work | Met | Multi-backend mode registers both backends; mt# prefix still routes to minsky backend |
| `getTaskSpecContentFromParams` uses multi-backend mode | Met | Passes `validParams.backend` (undefined when not provided) to `createConfiguredTaskService` — identical to other `*FromParams` functions |
| `migrate-backend` returns error details in non-JSON path | Met | `errorDetails: result.details?.filter((d: MigrationDetail) => d.error) ?? []` added to return value |
| Migration errors visible without `json: true` | Met | `errorDetails` array included in return regardless of format |

All spec criteria met.

PR #524, opened 2026-04-19
perf(mt#839): eliminate N+1 query patterns in subtask enrichment and graph traversal

edobry:
## Summary

- Add batch `getTasks(ids[])` to `TaskBackend` interface, implemented in Minsky DB (drizzle `inArray`), GitHub Issues (`Promise.all`), and multi-backend service (partition by prefix, batch per backend, fallback to sequential)
- Replace N `getTaskFromParams` calls in `crud-commands.ts` subtask enrichment with single `createConfiguredTaskService` + `getTasks(childIds)` — eliminates N+1 queries AND N service bootstraps
- Optimize `getAncestors` to prefetch parent edges via `getRelationshipsForTasks` and walk from memory
- Optimize `getTransitiveDependencies` to batch-fetch edges per BFS frontier instead of per-node `listDependencies` calls

## Test plan

- [ ] Existing tests pass (pre-commit hooks verify)
- [ ] `tasks get` with parent task still shows subtask summary correctly
- [ ] Multi-backend resolution (mt# parent with gh# children) still works
- [ ] Graph traversal cycle detection still functions correctly

### DIFF
--- a/src/adapters/shared/commands/tasks/crud-commands.ts
+++ b/src/adapters/shared/commands/tasks/crud-commands.ts
@@ -324,29 +324,42 @@ export class TasksGetCommand extends BaseTaskCommand<TasksGetParams> {
           const childIds = await service.listChildren(validatedTaskId);
 
           if (childIds.length > 0) {
+            const { createConfiguredTaskService } = await import(
+              "../../../../domain/tasks/taskService"
+            );
+            const { resolveRepoPath } = await import(
+              "../../../../domain/tasks/commands/shared-helpers"
+            );
+            const taskBaseParams = this.createTaskParams(params);
+            const workspacePath = await resolveRepoPath({
+              session: taskBaseParams.session as string | undefined,
+              repo: taskBaseParams.repo as string | undefined,
+            });
+            const childTaskService = await createConfiguredTaskService({
+              workspacePath,
+              backend: taskBaseParams.backend as string | undefined,
+              persistenceProvider: persistence,
+            });
+            const childTasks = await childTaskService.getTasks(childIds);
+
+            // Build a map for quick lookup; unresolved IDs fall back to "(unknown)"
+            const childTaskMap = new Map(childTasks.map((t) => [t.id, t]));
+
             const remaining: Array<{ id: string; title: string; status: string }> = [];
             let doneCount = 0;
             for (const childId of childIds) {
-              try {
-                const childTask = await getTaskFromParams(
-                  {
-                    ...this.createTaskParams(params),
-                    taskId: childId,
-                  },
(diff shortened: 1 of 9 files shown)


edobry (commented):
## Review: N+1 query elimination in subtask enrichment and graph traversal

**CI status:** in progress (checks running after force-push with fix)

### Findings

**[FIXED]** `crud-commands.ts:329` — Used `process.cwd()` for `workspacePath` instead of resolving via `resolveRepoPath({ session, repo })`. This would fail when `session` or `repo` params are set (e.g., inside a session workspace). Fixed in amended commit to use `resolveRepoPath` matching the original `getTaskFromParams` behavior.

### Checked and clear

- **`TaskBackend.getTasks?` interface** — correctly optional, won't break existing backends
- **`MinskyTaskBackend.getTasks`** — proper drizzle `inArray` query, empty-array guard
- **`GitHubIssuesTaskBackend.getTasks`** — `Promise.all` with null filtering, correct approach given no batch API
- **`TaskServiceImpl.getTasks`** — proper backend partitioning by prefix, fallback for unrouted IDs, fallback to sequential `getTask` if backend lacks `getTasks`
- **`FakeTaskService.getTasks`** — correct test double implementation
- **`getTransitiveDependencies`** — BFS frontier batching is correct; `getRelationshipsForTasks(frontier, "depends")` returns edges for all frontier nodes, and the `visited.has(edge.fromTaskId)` filter correctly follows only outgoing edges from the frontier
- **`getAncestors`** — functionally correct but marginal improvement (parent chains are linear, so each ancestor beyond the first still requires a separate query). Not a regression — same query count as before, just using `getRelationshipsForTasks` instead of `listFrom`
- **Test mock** — `tasks-core-functions.test.ts` correctly adds `getTasks` to satisfy the updated `TaskServiceInterface`

### Spec verification

**Task:** mt#839

|---|---|---|
| Subtask enrichment uses single batched operation | Met | `crud-commands.ts:326-340` — single `createConfiguredTaskService` + `getTasks(childIds)` replaces N `getTaskFromParams` calls |
| `getAncestors` uses efficient traversal | Met (marginal) | `task-graph-service.ts:239-264` — uses `getRelationshipsForTasks` with edge cache, but linear parent chains don't benefit much from batching |
| `getTransitiveDependencies` uses batched edge lookups | Met | `task-graph-service.ts:269-298` — frontier-based BFS with per-frontier batch `getRelationshipsForTasks` |
| No regression in multi-backend resolution | Met | `multi-backend-service.ts:229-274` — partitions by prefix, batch-queries each backend, handles unrouted IDs |

All spec criteria met. The `getAncestors` optimization is correct but provides marginal benefit for typical linear parent chains — acceptable given the spec says "batch or recursive CTE."

PR #565, opened 2026-04-20
perf(mt#878): optimize getAncestors to use single recursive CTE query

edobry:
## Summary

- Add `getAncestorChain(taskId, maxDepth)` to the `TaskRelationshipsRepository` interface
- Implement it in `createDrizzleRepo` using a single PostgreSQL recursive CTE, replacing the previous N-queries-for-N-ancestors loop
- Replace the `getAncestors` method body with a single delegation to `repo.getAncestorChain`
- Add an iterative in-memory implementation of `getAncestorChain` to the test fixture so all existing tests continue to pass

## Test plan

- [ ] Run `bun test --preload ./tests/setup.ts tests/domain/tasks/task-graph-service.test.ts` — all existing tests pass unchanged
- [ ] TypeScript compiles with 0 errors


edobry (commented):
## Review: Recursive CTE optimization for getAncestors

**CI status:** `Prevent Placeholder Tests` ✅ / `build` in progress

### Findings

No blocking or non-blocking issues found. The implementation is clean and correct.

### Checked and clear

- **CTE correctness**: Base case correctly selects `to_task_id` where `from_task_id = taskId AND type = 'parent'` (parent edges are child→parent). Recursive step correctly joins on `from_task_id = ancestor_id` to walk up the chain.
- **Depth limit**: `WHERE a.depth < ${maxDepth}` correctly caps recursion, matching the old loop's `for (let i = 0; i < maxDepth; i++)` semantics.
- **Result ordering**: `ORDER BY depth` ensures nearest-first ancestor ordering, matching previous behavior.
- **Drizzle `db.execute` return type**: `Array.from(result)` correctly handles the `RowList<Record<string, unknown>[]>` return type (which extends `Array`), avoiding the `as unknown as` cast that would trip the `no-excessive-as-unknown` lint rule.
- **`sql` import**: Already used elsewhere in the codebase (`task-relationships.ts`, `sqlite-storage.ts`), valid export from `drizzle-orm`.
- **In-memory test repo**: `getAncestorChain` implementation mirrors the old `getAncestors` loop logic faithfully — iterates up to `maxDepth`, walks parent edges, collects ancestors nearest-first.
- **No behavioral change**: `getAncestors` signature unchanged (`taskId: string, maxDepth = 10`), callers unaffected.

### Spec verification

**Task:** mt#878

| --- | --- | --- |
| `getAncestors` for N-deep task uses 1 DB query instead of N | Met | Single `db.execute(sql\`...\`)` call with recursive CTE replaces the N-iteration loop with per-iteration `getRelationshipsForTasks` calls |
| Cycle protection / max depth limit preserved | Met | `WHERE a.depth < ${maxDepth}` in CTE mirrors old `for (let i = 0; i < maxDepth; i++)` loop bound |
| Existing `task-graph-service.test.ts` tests pass unchanged | Met | Test file only adds `getAncestorChain` to in-memory repo; no test cases modified. Subagent confirmed all 1703 tests pass. |

### Documentation impact

No update needed — internal performance optimization with no user-facing or architectural changes.

PR #603, opened 2026-04-20
fix(mt#927): return pre-built message before lossy task formatter in formatResult

edobry:
## Summary

- `formatResult` in `base-task-command.ts` checked `"task" in result` before `"message" in result`, causing the lossy formatter (title + status only) to shadow the pre-built message that contains spec content, dep warnings, subtask info, etc.
- Added an early return at the top of the `"task"` branch: if `taskResult.message` is a non-empty string, return it directly.
- The generic task formatter is now a fallback for results that genuinely have no message.
- Removed the stale comment claiming spec content "needs to be fetched separately" — it is already in the message.

## Test plan

- [ ] Run `bun test --preload ./tests/setup.ts --timeout=15000 src/adapters/shared/commands/tasks/` and confirm all tests pass
- [ ] Verify that `tasks get`, `tasks edit`, and similar commands display full output (spec content, dep warnings) rather than just title + status

### DIFF
--- a/src/adapters/shared/commands/tasks/base-task-command.ts
+++ b/src/adapters/shared/commands/tasks/base-task-command.ts
@@ -145,6 +145,14 @@ export abstract class BaseTaskCommand<TParams = BaseTaskParams, TResult = unknow
       // Handle individual task details
       if (typeof result === "object" && result !== null && "task" in result) {
         const taskResult = result as Record<string, unknown>;
+
+        // Prefer the pre-built message — commands build it with full context
+        // (spec content, dep warnings, edit details, etc.) that the generic
+        // formatter below would lose.
+        if (taskResult.message && typeof taskResult.message === "string") {
+          return taskResult.message;
+        }
+
         const task = taskResult.task;
         if (task && typeof task === "object" && !Array.isArray(task)) {
           const { formatTaskIdForDisplay } = require("../../../../domain/tasks/task-id-utils");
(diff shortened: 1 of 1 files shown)


edobry (commented):
## Review: Fix formatResult message shadowing

**CI status:** ✅ All checks passed (build + Prevent Placeholder Tests)

### Findings

No blocking issues. The fix is minimal and correct.

### Checked and clear

- Early return logic: correctly checks for truthy string message before falling through to lossy formatter
- Empty message edge case: `""` is falsy, so falls through to structured formatter — correct behavior for any (hypothetical) result without a pre-built message
- Task list branch: unaffected — checked before this branch and doesn't use `createSuccessResult`
- JSON mode: unaffected — handled by `if (json) return result` at the top of `formatResult`

### Spec verification

**Task:** mt#927

| --- | --- | --- |
| `tasks_get` with `includeSpec: true` returns full spec content | Met | Message is built at crud-commands.ts:404 with `\n\nSpec:\n${extras.spec}`, now returned directly |
| Spec clearly separated and readable | Met | Message format uses double-newline separator |
| Without `includeSpec` still returns concise format | Met | Message is "Task X retrieved" without spec section |
| Subtask summary renders alongside spec | Met | Message is built with both (lines 394-404) |

### Documentation impact

No update needed — bugfix to internal formatter logic, no architectural or user-facing behavior change.

PR #607, opened 2026-04-20
fix(mt#939): thread sessionProvider through BaseGitOperation to GitService

edobry:
## Summary

- `BaseGitOperation.execute()` called `createGitService()` without passing the `sessionProvider` available in `this.deps`
- `GitService` then fell back to creating its own session provider via `createSessionProvider()` without persistence, causing the same "requires a persistence dependency" error on `session_commit` and `git_push`
- Now passes `sessionProvider` through to `createGitService()`, eliminating the fallback path
- `GitService.getSessionDb()` now throws a clear error instead of silently attempting an impossible operation

Follow-up to #605 (mt#936) — same class of error, different code path.

## Test plan

- [ ] `session_commit` works via MCP after restart
- [ ] `git_push` works via MCP after restart
- [ ] All unit tests pass (verified in pre-commit + pre-push hooks)

### DIFF
--- a/smoke-test.txt
+++ b/smoke-test.txt
@@ -1,1 +1,0 @@
-smoke test
--- a/smoke-test.txt
+++ b/smoke-test.txt
@@ -1,0 +1,1 @@
+smoke test
--- a/src/domain/git.ts
+++ b/src/domain/git.ts
@@ -3,11 +3,7 @@ import { join } from "node:path";
 import { mkdir } from "node:fs/promises";
 import { normalizeRepoName } from "./repo-utils";
 import { execAsync } from "../utils/exec";
-import {
-  createSessionProvider,
-  type SessionProviderInterface,
-  type SessionRecord,
-} from "./session";
+import { type SessionProviderInterface, type SessionRecord } from "./session";
 
 import { log } from "../utils/logger";
 import { getMinskyStateDir } from "../utils/paths";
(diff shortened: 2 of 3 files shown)


edobry (commented):
## Review: Thread sessionProvider through git operations

**CI status:** ✅ All checks passing

### Findings

No blocking issues. The fix correctly threads the available dependency through rather than letting GitService attempt to create its own.

### Checked and clear

- `BaseGitOperation` now passes `sessionProvider` to `createGitService()` — GitService will have it in `this.sessionDb` from construction
- The type widening in `GitOperationDependencies` is backward-compatible (added optional field)
- `GitService.getSessionDb()` now fails fast with a clear message instead of attempting an impossible operation
- No other callers of `createSessionProvider` with missing deps remain in non-test code (verified: only `composition/cli.ts:47` which correctly passes persistence)
- Removed import is genuinely unused after the fallback removal

### Spec verification

**Task:** mt#939

| --- | --- | --- |
| BaseGitOperation passes sessionProvider to createGitService() | Met | `base-git-operation.ts:65-67` |
| GitOperationDependencies type allows sessionProvider in factory | Met | `base-git-operation.ts:19-21` |
| GitService.getSessionDb() throws clear error | Met | `git.ts:112-114` |
| Unused import removed | Met | `git.ts:6` — only types imported now |
| session_commit works end-to-end via MCP | Pending | Requires MCP restart with this code |

### Documentation impact

No update needed — internal dependency threading fix, no architectural or behavioral changes.

PR #619, opened 2026-04-21
fix(mt#908): switch postgres migration execution from broken drizzle-kit subprocess to programmatic migrator

edobry:
## Summary

- Replace `runMigrationsWithDrizzleKit` (subprocess-based, silently fails) with `runSchemaMigrationsForConfiguredBackend` (programmatic, works correctly) in the postgres migration execution path
- Remove the now-unused `runMigrationsWithDrizzleKit` function (143 lines) from `migration-operations.ts`
- Remove the unused import of `runMigrationsWithDrizzleKit` from `persistence.ts`

## Changes

- `src/adapters/shared/commands/persistence.ts`: Replace broken drizzle-kit subprocess call with programmatic migrator; remove unused import
- `src/domain/persistence/migration-operations.ts`: Remove `runMigrationsWithDrizzleKit` function entirely since it has no remaining callers

## Test plan

- [ ] Run `bun test --preload ./tests/setup.ts --timeout=15000 src/domain/persistence tests/domain/persistence` to verify no regressions
- [ ] Verify that `persistence migrate` command with postgres backend now actually applies migrations instead of silently failing


edobry (commented):
## Review: Fix silent migration execution

**CI status:** `Prevent Placeholder Tests` ✅ / `build` in progress

### Findings

No blocking issues. The fix is minimal and correct.

### Checked and clear

- **Root cause addressed**: `runMigrationsWithDrizzleKit` spawned `bunx drizzle-kit migrate` which exited 0 without applying SQL. Replaced with `runSchemaMigrationsForConfiguredBackend` which uses drizzle-orm's programmatic `migrate()` — confirmed working during today's session.
- **Dead code removed**: Entire 147-line `runMigrationsWithDrizzleKit` function deleted. No other callers (verified via grep in diff).
- **Re-export preserved**: `getPostgresMigrationsStatus` re-export at line 32 kept — it uses the direct path import from `./postgres-migration-operations`, independent of the removed function.
- **`checkAndGenerateMigrations` preserved**: Uses drizzle-kit for `generate` and `check` (schema diffing), which is a different operation from migration execution and still works correctly.
- **Return type compatibility**: `runSchemaMigrationsForConfiguredBackend` returns `PostgresMigrationPlan | PostgresMigrationResult` which is a superset of the old `{ message, printed }` — the caller at line 146 just passes through `result`, so this is safe.

### Spec verification

**Task:** mt#908

| --- | --- | --- |
| `persistence migrate --execute` with pending migration applies it | Met | Now calls `runPostgresSchemaMigrations` which uses `await migrate(db, { migrationsFolder })` at postgres-migration-operations.ts:324 |
| After running, dry-run shows 0 pending | Met | Same migrator we verified manually today — applied 0018 successfully |
| Reports which migration(s) were applied | Met | postgres-migration-operations.ts:335 logs "Applied N migration(s) in Xms" with hash |
| If no pending, reports clearly | Met | postgres-migration-operations.ts lines 317-322 check pending count; also early-exit in dry-run path |
| If migration fails, error surfaced | Met | postgres-migration-operations.ts wraps in try/finally, drizzle-orm throws on SQL errors |

### Documentation impact

No update needed — internal bug fix, no user-facing behavioral change beyond "it works now."

